PR #13, opened 2021-11-02
URO-29 Added env variable support/defaults for PUBLIC_URL and REACT_APP_AUTH_SERIVCE_URL

dauglyon:
- adds a default `PUBLIC_URL` to `.env`
    - sidenote: this could be randomized during tests to help prevent accidental hard-coding of paths. @dakotablair thoughts?
- add `PUBLIC_URL` as the `react-router` `basename`, to ensure links work correctly
- add appropriate env variable values to the build action

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,8 @@
-EXTEND_ESLINT=true
+# Base URL path for the enviroment
+PUBLIC_URL = '/'
+
+# URL for the enviroment's Auth service
 REACT_APP_AUTH_SERIVCE_URL=https://ci.kbase.us/services/auth
+
+EXTEND_ESLINT=true
 SKIP_PREFLIGHT_CHECK=true
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 # Base URL path for the enviroment
-PUBLIC_URL = '/'
+PUBLIC_URL = '/dev/'
 
 # URL for the enviroment's Auth service
 REACT_APP_AUTH_SERIVCE_URL=https://ci.kbase.us/services/auth
--- a/.github/workflows/build_and_push.yml
+++ b/.github/workflows/build_and_push.yml
@@ -6,7 +6,7 @@ name: Build and Push to static-content
 on:
   # Triggers the workflow on push or pull request events but only for the main branch
   push:
-    branches: [ main ]
+    branches: [main]
   workflow_dispatch:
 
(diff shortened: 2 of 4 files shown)


.env
```diff
@@ -1,3 +1,8 @@
-EXTEND_ESLINT=true
+# Base URL path for the enviroment
+PUBLIC_URL = '/'
```
dauglyon, line 2: This is the path recommended for local dev

src/app/App.tsx
```diff
@@ -25,7 +25,7 @@ export default function App() {
   }, [token]);
 
   return (
-    <Router>
+    <Router basename={process.env.PUBLIC_URL}>
```
dauglyon, line 28: Makes `react-router` `Links` point to the right URLs 👍

.github/workflows/build_and_push.yml
```diff
@@ -27,6 +27,9 @@ jobs:
           node-version: '16'
 
       - name: Run build
+        env:
+          PUBLIC_URL: 'https://next-dev.kbase.us/services/static'
+          REACT_APP_AUTH_SERIVCE_URL: 'https://next-dev.kbase.us/services/auth'
         run: npm install && npm run build
```
dauglyon, line 33: pass the appropriate values during build

.env
dauglyon: We could set it to something different if we want to avoid the common pitfall of assuming everthing will be hosted at '/' during dev. 

e.g. if PUBLIC_URL is '/foo/bar' the local dev server root would be `localhost:3000/foo/bar/`

.github/workflows/build_and_push.yml
```diff
@@ -27,6 +27,9 @@ jobs:
           node-version: '16'
 
       - name: Run build
+        env:
+          PUBLIC_URL: 'https://next-dev.kbase.us/services/static'
```
dauglyon, line 31: should actually just be a path, without the domain.
```suggestion
          PUBLIC_URL: '/services/static'
```

.env
dakotablair: I agree, I've taken the liberty of making the default `/dev/` and making the `eslint`/`prettier` changes we previously discussed. Once the checks run again I will merge this PR.

dakotablair (approved):

PR #14, opened 2021-11-10
URO-27 Add TopBar and refactor /common/nav to /features/layout

dauglyon:
- Adds TopBar component
- Adds usePageTitle hook to set page and document title
- Modified storybook config to include redux state and router
- Changed index.html title

Needs thought:  
- Used redux to pass title & env. Use props instead?
- How to detect environment?
- usePageTitle hook is not compatible with class components


dakotablair (approved):
Looking good! Maybe we'll update the title based on what the user is looking at. That can be part of our next requirements meeting.

PR #15, opened 2021-11-16
URO-41: Add KBaseButton component.

dakotablair:
I am opening this PR early to discuss the general approach. There are no unit tests yet, so they will need to be added before this PR may be merged.


src/common/KBaseButton.module.scss
```diff
@@ -0,0 +1,11 @@
+.button {
+  /*button styles*/
+  border-color: #000;
```
dauglyon, line 3: Probably should use a kbase color from colors.scss (here and below)

dauglyon (commented):
Overall this looks good, though "KBaseButton" is a long name. Maybe we just call it "Button"? Also added some inline comments.

src/common/KBaseButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+  children,
+  disabled,
+  onclick,
+}) => {
+  return (
+    <button className={classes.button} onClick={onclick} disabled={disabled}>
+      {children}
+    </button>
+  );
+};
```
dauglyon, line 19:
```suggestion
export const KBaseButton: FC<React.HTMLProps<HTMLButtonElement>> = (props) => {
  return (
    <button {...props} className={[classes.button, props.className].join(' ')}>
      {props.children}
    </button>
  );
};
```

src/common/KBaseButton.tsx
```diff
@@ -0,0 +1,19 @@
+import { FunctionComponent as FC, MouseEventHandler } from 'react';
+import classes from './KBaseButton.module.scss';
+
+interface KBaseComponentProps {
+  disabled?: boolean;
+  onclick?: MouseEventHandler;
+}
```
dauglyon, line 7:
```suggestion
```

dauglyon (commented):
Also, the button props could just be recycled from `React.HTMLProps<HTMLButtonElement>` and then spread into the button. Custom props could be added later with `interface KbaseButtonProps extends React.HTMLProps<HTMLButtonElement> {someCustomeProp:type;}`

src/common/KBaseButton.module.scss
dakotablair: I might have gone too far in the other direction, but I think we'll want [classes like this](https://github.com/kbase/ui-refresh-test/pull/15/files#diff-6ea08c95a60c10153bc1d5ae1e330e1fa103cc61df85954a91f741359450a5b0R28) at some point anyway. Let me know what you think.

src/common/KBaseButton.tsx
dakotablair: This was almost perfect advice, but the type ended up being `FC<React.HTMLAttributes<HTMLButtonElement>>` after running into [this issue](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36505).

src/common/components/Button.module.scss
```diff
@@ -0,0 +1,24 @@
...
+
+  &[disabled] {
+    /*disabled button styles*/
+    background-color: use_color("neutral");
+    border-color: use_color("neutral");
+    color: use_color("silver");
+  }
+}
+
+@each $state, $_ in $palette {
```
dauglyon, line 15: scss power ftw

dauglyon (approved):
lgtm!

PR #16, opened 2021-11-17
URO-32 URO-34 TopBar Dropdown Menu

dauglyon:
## Still missing from this PR:
- ~~**Does not have functional menu items, they are just stubbed** this is due to not currently having access to external services. This either should be merged once we determine the appropriate behavior for each item, or we should generate a new JIRA task for filling these out. (This would also likely be the right time to add tests)~~
- storybook stories for dropdown and select

You might notice the `:global(.react-select__[name])` entries in the scss files. These are for react-select style overrides. If this is too arcane, we could also do this in JS using a different API they provide (which is less... weird). However, if we do a lot of customization of this component, it is probably best to have it in an .scss file. It's sort of a tradeoff...

For both select and dropdown, I forwarded props. This exposes all of the functionality of react-select, but might make it harder to excise if we choose a different dropdown lib in the future. So, that's another tradeoff to consider.

## Adds
- Adds react-select
- Adds New Select component using react-select
- Adds Dropdown component derived from the above Select component
- Uses Dropdown component to add Hamburger and Login menu functionality

- Screen Shot 2021-11-16 at 5 27 25 PM
- Screen Shot 2021-11-16 at 5 30 53 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "react-redux": "^7.2.4",
         "react-router-dom": "^5.2.0",
         "react-scripts": "4.0.3",
+        "react-select": "^5.2.1",
         "typescript": "^4.3.5",
         "web-vitals": "^1.1.2"
       },
@@ -1778,7 +1779,6 @@
     },
     "node_modules/@emotion/hash": {
       "version": "0.8.0",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/@emotion/is-prop-valid": {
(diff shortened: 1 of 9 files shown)


dakotablair:
Should the cursor change when hovering over the items in these menus since they are links?

dakotablair:
I think the only functionality we need with the menu items is that they link to the correct URLs, whether those URLs work is out of scope for these tickets.

dakotablair (approved):

PR #19, opened 2021-12-07
Fix stylelint and align scss to standard.

dakotablair:
This PR fixes the stylelint scripts to correctly lint all scss files. Some configuration is added to grandfather in current practice, but this can be revisited in a later configuration change. The biggest conceptual change is that `use_color` and `use_rgba_color` are now named `use-color` and `use-rgba-color` respectively.


dauglyon (approved):
lg(reat)tm

PR #21, opened 2021-12-07
URO-50 added line length setting

dauglyon:


dakotablair (approved):
In the coming days of the earth's magnetic field reversal our code will be able to be stored via punch card.

PR #22, opened 2021-12-11
Updating  REACT_APP_AUTH_SERIVCE_URL

jsfillman:
Updating to call CI for auth instead of next-dev, now that the content is running in CI.


dauglyon (approved):
LGTM, merging

PR #25, opened 2022-03-01
URO-46 Adds Async Search and Multi-select support to Select component

dauglyon:
Opted for adding text-select functionality into the existing Select component (used by the Dropdown component).
Bonus: we get multi-select

`Select` wraps react-select in a friendlier API suited to our needs and adds some styles. It has an on-change prop which takes a callback that receives selected options as an array on change. Text-search capability is always-on atm, but we could add a prop to disable it if we require that in the future.


src/common/components/Select.tsx
```diff
@@ -1,49 +1,130 @@
...
+      />
+    );
+  } else {
+    // Asynchrnous
+    // Could support pagination here if react-select-async-paginate was added
+    return (
+      <AsyncSelect
+        {...commonProps}
+        loadOptions={async (inputValue: string) => {
+          if (typeof props.options !== 'function') return [];
```
dauglyon, line 117:
```suggestion
          if (typeof props.options !== 'function') return props.options;
```

src/common/components/Select.tsx
```diff
@@ -1,49 +1,148 @@
...
+    isMulti: props.multiple,
+    onChange: handleChange,
+    value: props.value,
+  };
+
+  return !isFunc(options) ? (
+    // Synchrnous
+    <ReactSelect {...commonProps} options={options} />
+  ) : (
+    // Asynchrnous
```
dauglyon, line 120:
```suggestion
    // Synchronous
    <ReactSelect {...commonProps} options={options} />
  ) : (
    // Asynchronous
```

src/common/components/Select.tsx
```diff
@@ -1,49 +1,148 @@
...
+interface SelectProps {
+  /** Sets a className attribute on the outer component */
+  className?: string;
+  /** If true, adds a clickable icon for clearing the select */
+  clearable?: boolean;
+  /** Advanced: allows manualy manipulation of react-select subcomponents */
+  components?: ReactSelectProps<SelectOption>['components'];
+  /** If true, sets select to disabled */
+  disabled?: boolean;
+  /** Wheter the dropdown select menu should be pinned to the left or right
```
dauglyon, line 30:
```suggestion
  /** Whether the dropdown select menu should be pinned to the left or right
```

dakotablair (approved):
Looks good to me!

PR #26, opened 2022-03-29
Fix cross-origin links within nested iframes

dauglyon:


charleshtrenholm (approved):
Pretty dang impressive. One question - how would nested iFrames inside of iFrames be handled when theres CORS errors? Currently it looks like it will just fail within the iFrame

dauglyon:
It would be outside of the scope of this repo if that were the case, unless it was a CORS error during the link monkeypatching. But, it catches any CORS errors in nested iframes when monkeypatching and just ignores those frames.

PR #29, opened 2022-04-19
URO-80: Add coverage directory to eslintigorne.

dakotablair:
The `lint:strict` script was behaving poorly because it was attempting to lint the coverage artifacts which it should not do.


charleshtrenholm (approved):
Looks good

PR #32, opened 2022-05-03
URO-83 Add commit env variable and status page

dauglyon:


package.json
```diff
@@ -79,8 +79,8 @@
     }
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "REACT_APP_COMMIT=`git rev-parse HEAD` react-scripts start",
```
dauglyon, line 82: this could be pulled out into a separate ./scripts/start.sh and ./scripts/build.sh, but it's not needed unless we start adding more dynamic env variables...

charleshtrenholm (approved):
LGTM. We'll just add a script file if we ever need to add a second dynamic variable

PR #37, opened 2022-06-08
URO-94: icon redux slice

charleshtrenholm:
* Creates redux slice for handling synchronous and asynchronous icon state
* Create AppCellIcon component that handles async icons from user apps
* Creates TypeIcon component that loads predefined list of icons to display depending on KBase data type


src/common/components/TypeIcon.tsx
```diff
@@ -0,0 +1,86 @@
...
+  faFlask,
+  faCodeFork,
+  faBullseye,
+  faArrowsLeftRight,
+  faGaugeHigh,
+  faBook,
+  faFile,
+} from '@fortawesome/free-solid-svg-icons';
+
+library.add(
```
dauglyon, line 26: This step seems like it should happen in a common module shared by TypeIcon, AppCellIcon, and any other place we might use icons (LeftNavBar?). Not clear on exactly what it is doing. Does this make the icons accessible via string name?

dauglyon: could maybe export a wrapped FAIcon which extends the icon prop to accept a string? (see typegaurd comment above)

src/common/components/kbase_icons.css
```diff
@@ -0,0 +1,58 @@
+@font-face {
+  font-family: "kbase-icons";
+  src:url("https://ci.kbase.us/ui-assets/fonts/v1/kbase-icons.eot");
+  src:url("https://ci.kbase.us/ui-assets/fonts/v1/kbase-icons.eot?#iefix") format("embedded-opentype"),
+      url("https://ci.kbase.us/ui-assets/fonts/v1/kbase-icons.woff") format("woff"),
+      url("https://ci.kbase.us/ui-assets/fonts/v1/kbase-icons.ttf") format("truetype"),
+      url("https://ci.kbase.us/ui-assets/fonts/v1/kbase-icons.svg#kbase-icons") format("svg");
```
dauglyon, line 7: do we know how https://github.com/kbase/ui-assets is deployed? Should these files instead be in `/public`? 

this is probably fine for now, but I imagine we'll want to merge ui-assets into europa asap? @charleshtrenholm @dakotablair thoughts on that?

src/common/components/kbase_icons.css
```diff
@@ -0,0 +1,58 @@
...
+  content: attr(data-icon);
+  font-style: normal !important;
+  font-weight: normal !important;
+  font-variant: normal !important;
+  text-transform: none !important;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+[class^="icon-"]:before,
```
dauglyon, line 23: global css sort of freaks me out. maybe we should have a kbaseIcon component so this can be a css module? thoughts?

src/common/slices/iconSlice.ts
```diff
@@ -0,0 +1,121 @@
+import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { getServiceClient, URLS } from '../services';
+import { typeIconInfos } from './icons';
+
+export interface IconInfo {
+  icon?: string;
+  color: string;
+  url?: string;
+  isImage: boolean;
+}
```
dauglyon, line 10: this type seems a little strange. Maybe consider the following to prevent having to check that icon/url are defined after checking isImage (as, if i'm reading this correctly, `isImage` tells you which should be defined)

```suggestion
export type IconInfo = {
    isImage: true;
    url: string;
    color: string;
} | {
    isImage: false;
    icon: string;
    color: string;
};
```

src/common/slices/iconSlice.ts
```diff
@@ -0,0 +1,121 @@
...
+  appIconCache: AppIconCache;
+  defaultApp: IconInfo;
+  defaultType: IconInfo;
+  loadingIcon: IconInfo;
+}
+
+type AppIconPayload = {
+  appId: string;
+  appTag: AppTag;
+  icon?: IconInfo;
```
dauglyon, line 34: Probably being pedantic but... in what situation is this undefined? if it's that one is the input for the thunk, and one is the output, consider using two different types

src/common/components/AppCellIcon.tsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+  // anything thats not an image will be the loading icon or the default icon
+  return (
+    <div
+      className={classes.loaded_icon_outer}
+      style={{ backgroundColor: icon.color }}
+    >
+      <FAIcon
+        className={classes.loaded_icon_inner}
+        icon={icon.icon as IconProp}
```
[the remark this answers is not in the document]
dauglyon: (this is somewhat pedantic but)

dauglyon (commented):
Overall looks good, made some minor comments. The only one keeping me from hitting approve is the one about pulling the fontawesome initialization code out into its own file/module

src/common/components/AppCellIcon.tsx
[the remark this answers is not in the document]
charleshtrenholm: Ahh, that was from before I upgraded the fontawesome libraries. It was a type bug that ended up fixing itself after moving from v1 to v6

src/common/components/AppCellIcon.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  expect(container.querySelector('svg[data-icon="spinner"]')).not.toBeNull();
+});
+
+test('AppCellIcon renders an icon after the callback finishes', async () => {
+  const { container } = render(
+    <Provider store={createTestStore()}>
+      <AppCellIcon appId="SomeModule.someApp" appTag={AppTag.beta} />
+    </Provider>
+  );
+  await waitFor(() => {
```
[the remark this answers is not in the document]
charleshtrenholm: Huh. The test was working fine for me. I can try again

src/common/components/kbase_icons.css
charleshtrenholm: This was pulled in from navigator. I tried importing it first into the css module but it wasn't working for some reason. I can try again

src/common/slices/iconSlice.ts
charleshtrenholm: When I try to do this, trypescript blows up because its expecting one type or another in different contexts.

src/common/components/AppCellIcon.tsx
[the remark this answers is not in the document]
dauglyon: is this assertion still needed given the new changes?

src/stories/components/AppCellIcon.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+        <AppCellIcon {...args}></AppCellIcon>
+      </div>
+    </Router>
+  );
+};
+
+export const Default = AppCellIconTemplate.bind({});
+
+Default.args = {
+  appTag: 'release' as AppTag,
```
dauglyon, line 28: this should probably be an enum member instead of an assertion ie "AppTag.release"

dauglyon (approved):
Lgtm, just left one minor comment above

PR #38, opened 2022-06-10
URO-88 dataview component

charleshtrenholm:
Creates DataView layout component with icons and links to narrative data types.


src/common/components/DataView.tsx
```diff
@@ -0,0 +1,60 @@
+import { DataObject } from '../types/NarrativeDoc';
```
dauglyon, line 1: Would recommend moving this file and other DataView files into a DataView folder (like NarrativeList)

src/common/components/DataView.module.scss
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.dataview a,
+.dataview a:visited {
+  color: #000;
+  text-decoration: none;
+}
+
+.dataview_row_type {
+  color: rgba(0, 0, 0, 0.8);
```
dauglyon, line 27: For colors in this file, could they be references to colors in `src/common/colors.scss` (or could they be changed to be one of those colors?).

src/common/components/DataView.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const DataView: FC<DataViewProps> = ({ accessGroup, dataObjects }) => {
+  if (!dataObjects?.length) {
+    return (
+      <p className={classes.dataview_no_data}>This narrative has no data.</p>
+    );
+  }
+
+  const rows = dataObjects
+    .slice(0, 50)
```
dauglyon, line 20: Is there any indication to the user that this list gets truncated? Might be something to create a future ticket to address, if it isn't trivial to add.

src/common/components/DataView.tsx
```diff
@@ -0,0 +1,60 @@
+import { DataObject } from '../types/NarrativeDoc';
+import { FC } from 'react';
+import { getWSTypeName } from '../utils/stringUtils';
+import classes from './DataView.module.scss';
+import TypeIcon from './TypeIcon';
+
+interface DataViewProps {
+  accessGroup: number;
```
dauglyon, line 8: I'm not clear on what accessGroup is at first reading. Is there a clearer name for this prop? Or maybe a comment explaining what it is?

src/common/utils/stringUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Make a readable string from a snake-case string
+ * Eg. "comparative_genomics" -> "Comparative genomics"
+ * This makes the string sentence case. E.g.:
+ * some_kind_of_string -> Some kind of string
+ * @param {string} str the snake-case string to change
+ * @return {string}
+ */
+export function formatSnakeCase(str: string): string {
```
dauglyon, line 13: nit: this function actually _parses_ snake case

dauglyon: maybe snakeCaseToWords or snakeCaseToHumanReadable or something

src/common/components/DataView.tsx
charleshtrenholm: I think a future ticket would likely be the best scenario. This was the UI pulled verbatim from navigator. I agree that it should show some indication that it's not displaying all the objects but I'm not sure I have a design choice for that off the top of my head that would satisfy everyone. Since it likely involves further discussion I think it would be lest to leave as is in this particular PR

dauglyon (approved):

PR #39, opened 2022-06-22
URO-52, URO-65 Add profile slice and component.

dakotablair:
URO-52 Profile component: This commit introduces <ProfileWrapper /> to manage state of profile view. The <Profile /> component is mostly a template and holds very little logic, but handles passing the profile data to the profile view and
the narrative view.

URO-65 Profile slice: The new profile slice holds all data which is visible in the profile view. This commit introduces an ad-hoc caching layer for profile data. It retrieves data from the profile service using @kbase/narrative-utils.

This commit also includes the Meta type to avoid explicitly typing portions of service call responses not currently implemented. It factors <Routes /> into its own module and adds jest-fetch-mock for testing HTTP requests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "@fortawesome/free-solid-svg-icons": "^6.1.1",
     "@fortawesome/react-fontawesome": "^0.1.18",
     "@kbase/narrative-utils": "^0.0.6",
-    "@redux-devtools/extension": "^3.2.2",
     "@reduxjs/toolkit": "^1.6.1",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^11.2.7",
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@fortawesome/free-solid-svg-icons": "^6.1.1",
     "@fortawesome/react-fontawesome": "^0.1.18",
     "@kbase/narrative-utils": "^0.0.6",
+    "@redux-devtools/extension": "^3.2.2",
     "@reduxjs/toolkit": "^1.6.1",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^11.2.7",
@@ -26,6 +27,7 @@
     "@types/node": "^12.20.17",
     "@types/react": "^17.0.15",
     "@types/react-dom": "^17.0.9",
+    "jest-fetch-mock": "^3.0.3",
     "node-sass": "^6.0.1",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
@@ -61,6 +63,9 @@
   },
   "engines": {
     "node": ">=16.13.0"
+  },
+  "jest": {
+    "resetMocks": false
   },
   "name": "ui-refresh-test",
   "private": true,
--- a/src/app/store.ts
+++ b/src/app/store.ts
@@ -1,24 +1,29 @@
 import { configureStore } from '@reduxjs/toolkit';
-import type { Meta } from '../common/types';
 import auth from '../features/auth/authSlice';
 import count from '../features/count/countSlice';
 import icons from '../common/slices/iconSlice';
 import layout from '../features/layout/layoutSlice';
 import profile from '../features/profile/profileSlice';
 
-const createStore = (preloadedState?: Meta) => {
+const createStore = () => {
   const config = {
     devTools: true,
-    preloadedState: preloadedState ? preloadedState : {},
     reducer: { auth, count, icons, layout, profile },
   };
   return configureStore(config);
 };
 
 export const store = createStore();
-export const createTestStore = (preloadedState?: Meta) =>
-  createStore(preloadedState);
 
 // Infer the `RootState` and `AppDispatch` types from the store itself
 export type RootState = ReturnType<typeof store.getState>;
 export type AppDispatch = typeof store.dispatch;
+
+export const createTestStore = (preloadedState: Partial<RootState>) => {
+  const config = {
+    devTools: true,
+    preloadedState: preloadedState,
+    reducer: { auth, count, icons, layout, profile },
+  };
+  return configureStore(config);
+};
--- a/src/app/store.ts
+++ b/src/app/store.ts
@@ -1,16 +1,23 @@
 import { configureStore } from '@reduxjs/toolkit';
-import count from '../features/count/countSlice';
+import type { Meta } from '../common/types';
 import auth from '../features/auth/authSlice';
-import layout from '../features/layout/layoutSlice';
+import count from '../features/count/countSlice';
 import icons from '../common/slices/iconSlice';
+import layout from '../features/layout/layoutSlice';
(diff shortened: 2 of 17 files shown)


src/app/store.ts
```diff
@@ -1,16 +1,23 @@
...
+import profile from '../features/profile/profileSlice';
 
-const createStore = () =>
-  configureStore({
-    reducer: { count, auth, layout, icons },
-  });
+const createStore = (preloadedState?: Meta) => {
+  const config = {
+    devTools: true,
+    preloadedState: preloadedState ? preloadedState : {},
```
charleshtrenholm, line 12: Maybe it would be neater if the `preloadedState` var was moved completely out of `createStore` and having `createTestStore` be the function that could handle preloaded state

src/features/profile/Profile.module.scss
```diff
@@ -0,0 +1,23 @@
+.narratives {
```
charleshtrenholm, line 1: I think in order to complete the PR the CSS should be completed to an approximation of the current css. At least the dashed lines around the elements should be removed.

src/features/profile/Profile.tsx
```diff
@@ -0,0 +1,186 @@
...
+import classes from './Profile.module.scss';
+
+export const ProfileNarrativesMessage: FC<{
+  realname: string;
+  yours: boolean;
+}> = ({ realname, yours }) => {
+  if (yours) {
+    return <span>This table shows all of your narratives.</span>;
+  }
+  return (
```
charleshtrenholm, line 29: I think that all of these small components (`ProfileNarrativeMessage`, `ProfileResume`, `ProfileInfobox`, etc.) should be folded into the `Profile` component. when you just have nested components inside of components that effectively render only DOM elements, you lose any of the advantages of breaking things down into subcomponents and start obfuscating what the actual structure of the html itself is. I just think its easier for developers down the road to follow the HTML

src/features/profile/Profile.tsx
```diff
@@ -0,0 +1,186 @@
...
+}) => {
+  return (
+    <div className={classes.narratives}>
+      <ProfileNarrativesMessage realname={realname} yours={yours} />
+      <NarrativeList
+        items={[]}
+        itemsRemaining={0}
+        hasMoreItems={false}
+        loading={false}
+        onLoadMoreItems={() => {}} //eslint-disable-line @typescript-eslint/no-empty-function
```
charleshtrenholm, line 53: `onLoadMoreItems` and `onSelectItem` were made deliberately optional in case they didn't need to be used, so you don't need to add the empty functions and the eslint disables.

src/features/profile/Profile.tsx
```diff
@@ -0,0 +1,186 @@
...
+        ) : (
+          <ProfileView realname={realname} />
+        )}
+        <Button>Edit</Button>
+      </section>
+    </>
+  );
+};
+
+export const ProfileWrapper: FC = () => {
```
charleshtrenholm, line 118: I like this wrapper that handles all the redux logic, it makes the storybook display really straightforward. I was having trouble doing something like this within a storybook component for one of the navigator items.

src/features/profile/Profile.tsx
```diff
@@ -0,0 +1,186 @@
...
+  }
+  const loadUsername = usernameNotSpecified ? usernameAuthed : viewUsername;
+  if (!(loadUsername in cachedProfiles)) {
+    dispatch(loadProfile({ token: tokenString, username: loadUsername }));
+  }
+  const profile = cachedProfiles[loadUsername];
+  if (!profile) {
+    return <>Loading user profile.</>;
+  }
+  const profileNames = cachedProfiles[loadUsername].user;
```
charleshtrenholm, line 165: This is pretty small, but can't `cachedProfiles[loadUsername].user` just be `profile.user` since you checked that it's defined?

src/features/profile/profileSlice.ts
```diff
@@ -0,0 +1,161 @@
...
+      })
+      .addCase(loadProfile.fulfilled, (state, action) => {
+        const { profile: profileCache } =
+          action.payload as KBaseProfileResponseCache;
+        const { user } = profileCache;
+        state.realname = user.realname;
+        state.usernameSelected = user.username;
+        const cache: KBaseProfileResponseCache = {
+          [state.usernameSelected]: profileCache,
+        };
```
charleshtrenholm, line 130: Wouldn't this only allow one profile to be cached at a time? I'm assuming you want to cache more than one. Wouldn't you want the logic here to be something more like this?:
```typescript
const cache: KBaseProfileResponseCache = {
  [state.usernameSelected]: profileCache,
  ...state.cache
}
```

charleshtrenholm: Its possible that you'd only want to cache one at a time, but if that's true then I'm not sure why `cache` would need to be a key value store

charleshtrenholm (changes requested):
Looks good, a few changes requested. Namely taking a look at the ProfileSlice issue and working on the CSS for profile

src/features/profile/Profile.tsx
dakotablair: Thanks! I will make this update in the next iteration.

src/features/profile/Profile.tsx
dakotablair: Thanks to @dauglyon for the suggestion!

package.json
```diff
@@ -18,6 +18,7 @@
     "@fortawesome/free-solid-svg-icons": "^6.1.1",
     "@fortawesome/react-fontawesome": "^0.1.18",
     "@kbase/narrative-utils": "^0.0.6",
+    "@redux-devtools/extension": "^3.2.2",
```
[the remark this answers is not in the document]
dakotablair: I was using this, but am not anymore. I will remove it.

src/features/profile/Profile.module.scss
dakotablair: We talked about this and made a decision as a team: [URO-99](https://kbase-jira.atlassian.net/browse/URO-99)

src/app/store.ts
dakotablair: This change is included in the latest commit.

package.json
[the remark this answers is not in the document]
dakotablair: Removed in the latest commit.

src/features/profile/Profile.tsx
dakotablair: Added an explanatory comment in the latest commit.

src/features/profile/Profile.tsx
dakotablair: Updated in the latest commit.

src/features/profile/profileSlice.ts
dakotablair: Fixed in the latest commit.

charleshtrenholm (approved):
LGTM!

PR #46, opened 2022-08-03
URO-35 rtk query api

dauglyon:
Uses rtk-query to create API hooks, includes auth header formatting, jsonrpc logic (including checking the id), and dynamic service // service-wizard support. My main concern is that it is "over fit" to rtk-query, and if rtk-query falls out of fashion, it may be hard to swap it out in-place. Though, likely easier than a hand-rolled solution? After going back and forth while implementing this, I've landed on really liking rtk-query.

The bulk of the custom logic (auth header, service wizard usage) lives in `src/common/api/utils/kbaseBaseQuery.ts`. 
There is a usage example in `src/features/auth/Auth.tsx`.


src/common/api/index.ts
```diff
@@ -0,0 +1,11 @@
+import { kbaseBaseQuery } from './utils/kbaseBaseQuery';
+import { createApi } from '@reduxjs/toolkit/query/react';
+
+export const baseApi = createApi({
+  reducerPath: 'combinedApi',
+  baseQuery: kbaseBaseQuery({ baseUrl: 'https://ci.kbase.us/' }),
```
dauglyon, line 6: should be based on the enviroment

src/common/api/wsObjectApi.ts
```diff
@@ -0,0 +1,35 @@
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react';
```
dauglyon, line 1: need to refactor this file, right now it wont work because it isnt hooked up to state (as it isnt consuming the baseAPI)

src/common/api/utils/kbaseBaseQuery.ts
```diff
@@ -0,0 +1,183 @@
...
+      if (isDynamic(kbQueryArgs.service)) {
+        // This api is provided via the query to prevent circular imports, fun
+        const serviceStatusQuery =
+          getConsumedService('serviceWizardApi').endpoints.serviceStatus;
+        const wizardQueryArgs = {
+          module_name: kbQueryArgs.service.name,
+          version: kbQueryArgs.service.release,
+        };
+
+        // trigger query, subscribing until we grab the value
```
dauglyon, line 88: add more comments here, we could make this refetch number larger

dakotablair (approved):
Looking good in the neighborhood.

PR #49, opened 2022-09-19
Refactor exisiting service usage to RTK-Query

dauglyon:
Refactored all existing service calls to go through rtk-query. Because it has a slice, as well as components, I also moved Icons to a feature folder.


.env
```diff
@@ -2,7 +2,6 @@
 PUBLIC_URL = '/dev/'
 
 # URL for the enviroment's Auth service
```
dauglyon, line 4:
```suggestion

```

src/common/api/authService.ts
```diff
@@ -0,0 +1,72 @@
...
+  created: number;
+  expires: number;
+  id: string;
+  name: string | null;
+  type: string;
+  user: string;
+  cachefor: number;
+}
+
+// Auth does not use JSONRpc, so we use queryFn to make custom queries
```
dauglyon, line 16: When I start collections I will set up our default baseQuery so it can handle simple http calls alongside jsonRpc

src/common/api/userProfileApi.ts
```diff
@@ -33,7 +33,7 @@ export const userProfileApi = baseApi
   .enhanceEndpoints({ addTagTypes: ['Profile'] })
   .injectEndpoints({
     endpoints: (builder) => ({
-      status: builder.query<
+      userProfileStatus: builder.query<
```
dauglyon, line 36: we cant have duplicate endpoint names across APIs, hence this change

src/common/api/utils/parseError.ts
```diff
@@ -0,0 +1,40 @@
+import { SerializedError } from '@reduxjs/toolkit';
+import { KBaseBaseQueryError } from './kbaseBaseQuery';
+
```
dauglyon, line 3:
```suggestion

// TODO: Add Tests
```

src/common/services.ts
```diff
@@ -1,20 +0,0 @@
-import { KBaseServiceClient } from '@kbase/narrative-utils';
```
dauglyon, line 1: `@kbase/narrative-utils` needs to be removed

dakotablair (approved):
This is already a big lift, so I am glad we are doing it now rather than later. Looks good!

PR #51, opened 2022-10-06
URO-38 Add text input component.

dakotablair:
This PR creates a KBase text input react component and adds an example story to storybook. To properly test this component, this commit upgrades the `@testing-library/user-event` library. This upgrade necessitates some fixes to tests in the Select component, including some slight refactoring and new tests to bring it up to 100% coverage.

- [x] Creates a KBase text input react component.
- [x] Adds a storybook story demonstrating text input.
- [x] Includes validation error state display.
- [x] Includes disabled state.
- [x] Does not violate principles of Europa.


src/common/components/Input.tsx
```diff
@@ -0,0 +1,81 @@
...
+  ) {}
+}
+
+const InputInterfaceKeys = Object.keys(new InputInterfaceClass('', <></>));
+
+interface InputInterface
+  extends React.ComponentProps<'input'>,
+    InputInterfaceClass {}
+
+export const Input: FC<InputInterface> = (props) => {
```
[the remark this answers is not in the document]
dauglyon: Component which requires exclusively _either_ vaildate _or_ onChange

```tsx
interface InputInterface extends Omit<React.ComponentProps<'input'>, 'onChange'> {
  idForLabel: string; // should probably be auto-generated
  label: ReactElement; // could be optional
}

interface ValidatorInputInterface extends InputInterface {
  validate: (value: string) => boolean;
  onChange?: undefined;
}
interface ChangeInputInterface extends InputInterface {
  validate?: undefined;
  onChange: React.ComponentProps<'input'>['onChange'];
}

export const Input: FC<ValidatorInputInterface | ChangeInputInterface> = (
  props
) => {/*...*/}
```

src/common/components/Input.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <label className={labelClass} htmlFor={idForLabel}>
+        {label}
+      </label>
+      <input
+        className={inputClass}
+        id={idForLabel}
+        onChange={handler}
+        type={'text'}
+        {...otherProps}
+      />
```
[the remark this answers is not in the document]
dauglyon: This also lets use the usual `onChange` as a prop, as we overwrite the value in `props` by explicitly setting it after the spread.

src/common/components/Select.tsx
```diff
@@ -41,6 +41,20 @@ interface SelectProps {
   value?: SingleValue<SelectOption> | MultiValue<SelectOption>;
 }
 
+export const handleChangeFactory = (
```
dauglyon, line 44: It might be worth coming up with a prefix we can use for exports which are only meant to be consumed by tests. The pythonista in my thinks `export const _handleChangeFactory` could work well.

src/common/components/Input.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    consoleError.mockClear();
+  });
+
+  test('Input Renders', () => {
+    render(<Input idForLabel={'test'} label={<>test</>} />);
+    const inputText = screen.getByLabelText('test', { exact: false });
+    expect(inputText).toBeInTheDocument();
+  });
+
+  test('Specifying changeHandler and validate throws an error', () => {
```
dauglyon, line 31: This is a neat testing strategy, but I think more precisely typing <Input/> will eliminate the need for it here. (Sorry 🙈 I'm only just now understanding what you meant about errors inside the component)

src/common/components/Input.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+      </Provider>
+    );
+    const errorText = screen.getByText(handlerPropError, { exact: false });
+    expect(errorText).toBeInTheDocument();
+  });
+
+  test('Specifying validate validates input', async () => {
+    const validate = (value: string) => value === 'taco';
+    render(
+      <Input idForLabel={'test'} label={<>Validate</>} validate={validate} />
```
dauglyon, line 54: How does one get the value of a validated input from its parent component?

src/common/components/Input.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    });
+    const inputElement = screen.getByLabelText('Validate', {
+      exact: false,
+    }) as HTMLInputElement;
+    Array.from(document.getElementsByTagName('label')).forEach(
+      (label: HTMLLabelElement, ix: number) => {
+        label.remove();
+      }
+    );
+    await userEvent.type(inputElement, 'a');
```
dauglyon, line 99: Not clear on why this user action throws an error...

dauglyon (changes requested):
A selection of comments, most around the relationship between changeHandler and validator, and around controlled input support.

src/common/components/Input.common.tsx
```diff
@@ -0,0 +1,6 @@
+export const missingLabelFactory = (id: string) => {
+  return `Could not find the label for #${id}, has it been removed?`;
+};
+
+export const handlerPropError =
```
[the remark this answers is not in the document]
dakotablair: This concern is mooted by using `react-hook-form`.

src/common/components/Input.tsx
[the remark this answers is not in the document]
dakotablair: Great points, but using `react-hook-form` makes the code even simpler.

src/common/components/Input.tsx
```diff
@@ -0,0 +1,81 @@
...
+    InputInterfaceClass {}
+
+export const Input: FC<InputInterface> = (props) => {
+  const [error, setError] = useState('');
+  const { className, changeHandler, idForLabel, label, validate } = props;
+  const otherProps = Object.fromEntries(
+    Object.entries(props).filter(
+      ([key, value]) => InputInterfaceKeys.indexOf(key) === -1
+    )
+  );
```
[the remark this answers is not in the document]
dakotablair: This code can be avoided by [appropriately ordering the passed props](https://github.com/kbase/ui-refresh-test/pull/51/files#diff-dfef91aeb90414e4680f304a5a3d1e0557825a7d9d0f59af5dfc8cb68e70ab03R49-R55).

src/common/components/Input.test.tsx
dakotablair: Using `react-hook-form` made the tests simpler and dispensed with the need for this kind of error handling at this time. We might use this pattern at some other point in a separate context where errors must be tested.

src/common/components/Input.tsx
```diff
@@ -0,0 +1,60 @@
...
+      validated ? statusClass : '',
+      className,
+    ].join(' ');
+    let handler = onChange;
+    if (onChangeAlso && onChange) {
+      handler = (event) => {
+        onChangeAlso(event);
+        onChange(event);
+      };
+    }
```
dauglyon, line 42: Just in case the onChange is "used" but not truthy, so that onChangeAlso doesn't mysteriously not work, I recommend maybe:
```suggestion
    let handler = onChange || onChangeAlso;
    if (onChangeAlso && onChange) {
      handler = (event) => {
        onChangeAlso(event);
        onChange(event);
      };
    }
```

dauglyon: alternatively just:
```suggestion
      handler = (event) => {
        if(onChangeAlso) onChangeAlso(event);
        if(onChange) onChange(event);
      };
```

src/common/components/Input.common.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const classNames = Object.keys(classes);
+  const randomIndex = Math.floor(classNames.length * Math.random());
+  const randomClass = classes[classNames[randomIndex]];
+  const currentClasses = evt.currentTarget.classList;
+  currentClasses.forEach((cls) => currentClasses.remove(cls));
+  currentClasses.add(classes.input);
+  currentClasses.add(randomClass);
+};
+
+export const inputRegisterFactory = <T,>({
```
dauglyon, line 21: I'm still not sure how I feel about this, it seems... hacky

dauglyon: Looking at https://react-hook-form.com/api/useform, validation occurs by default on onChange events, which means it will only validate the contents of a form if they are changed. That seems like preferred behavior over this perhaps? I'm not clear what benefit we get from this

src/common/components/Input.tsx
dauglyon: Further, it looks like `register` can take an optional "onChange" that may eliminate our need for onChangeAlso, assuming we don't have other places where onChangeAlso is needed alongside onChange

dauglyon: https://react-hook-form.com/api/useform/register

dauglyon (commented):
Overall looks good, with a few questions/suggestions

src/common/components/Input.tsx
```diff
@@ -0,0 +1,60 @@
...
+export const Input = forwardRef<HTMLInputElement, InputInterface>(
+  (props, ref) => {
+    const { className, errors, onChangeAlso, validated, label, ...rest } =
+      props;
+    const { name, onChange } = rest; // react-hook-form internals
+    const idForLabel = useMemo(() => `input-${uuidv4()}`, []);
+    const statusClass = errors ? classes.error : classes.success;
+    const inputClass = [
+      classes.input,
+      validated ? statusClass : '',
```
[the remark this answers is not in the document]
dakotablair: The pseudo-classes would be used in the CSS. We would need `react-hook-form` to make more use of the [Constraint validation API](https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation) to make use of them. In particular, the `:invalid` pseudo-class (by default? I am not sure if this is configurable) is not added when the validations fail. You can observe this by adding an `:invalid` rule in the stories and entering an invalid value. For simple validations (max/min length, pattern) we could do a little extra work and pass them along to the input tag, but I am not sure it is worth it to only get half the functionality.

src/common/components/Input.tsx
dakotablair: Yes, I think that will work for the use case `onChangeAlso` was meaning to address, that is, other side effects of user change events not related to validation. The main difference is that instead of it being a prop to `<Input />` it's a parameter in the register options. But users may still have a question of which fires first, `onChange` or validation, and it appears for now it is validation and then `onChange` as seen [here in the repo](https://github.com/react-hook-form/react-hook-form/blob/6ad4da33dcea2584976aeb9d1f0c1c2bc6803000/src/logic/createFormControl.ts#L647-L681).

src/common/components/Input.common.tsx
dakotablair: The `validated` prop is meant to indicate whether or not the input has been validated and the errors are the errors for all the fields. The errors are only passed to a component if there is an error for that named field. The alternative to this is to do this logic when props are passed to the <Input /> component, but you want to do the same thing each time, so I put it in the factory to keep things DRY.

dauglyon (approved):

PR #55, opened 2022-11-01
URO-132 Add Card and CardList components

dauglyon:


src/common/components/Card.module.scss
```diff
@@ -0,0 +1,95 @@
...
+  flex-flow: column nowrap;
+  flex-grow: 1;
+  gap: 5px;
+  justify-content: center;
+
+  .title > & {
+    flex-grow: 0;
+  }
+
+  .subtitle {
```
dauglyon, line 65:
```suggestion
  justify-content: center;

  .subtitle {
```

package.json
```diff
@@ -32,6 +32,7 @@
     "react-error-boundary": "^3.1.4",
     "react-hook-form": "^7.37.0",
     "react-redux": "^7.2.4",
+    "react-router": "^6.4.3",
```
dauglyon, line 35: This is incorrect, need to make a new commit to fix it

dakotablair (approved):
- [x] Component is responsive to its container size
- [x] Component accepts the following props as react elements and positions them appropriately with:
  - [x] title
- [x] Component should be displayed according to the context provided by its props. Additional props include:
  - [x] linkTo
  - [x] selected

PR #61, opened 2023-01-05
RE2022-41 Add support for dynamic table data

dauglyon:
Adds typing and props to allow for the Table component to be used with dynamic (server-side sorted/paginated) data.

Also fixes a bug where tables with zero-length data arrays (zero rows) would crash during render.


src/common/components/Table.tsx
```diff
@@ -37,14 +64,12 @@ export const Table = <Datum extends unknown>({
...
   className,
-  pageSize = false,
-}: {
-  data: Datum[];
-  columnDefs?: ColumnDefs<Datum>;
-  className?: string;
-  pageSize?: number | false;
-}) => {
-  const [sorting, setSorting] = useState<SortingState>([]);
+  pageSize = Number.MAX_SAFE_INTEGER,
```
dakotablair, line 67: This seems like an impractical default limit.

src/common/components/Table.tsx
```diff
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
```
dakotablair, line 1: Note to Future David: Have brain thoughts about more documentation and more intuitive api.

PR #71, opened 2023-03-29
DEVOPS-1222 Update root URL for collections service.

dakotablair:
- [ ] Change the [collection service path](https://github.com/kbase/ui-refresh-test/blob/main/src/common/api/collectionsApi.ts#L7) from `collectionsservice` to `collections`.


dakotablair:
See PR #66.

PR #72, opened 2023-04-12
RE2022-122 Add selection pane, api calls, and basic state management

dauglyon:
does not include any selection UI within data products


dakotablair (approved):
Looks good to me!

PR #76, opened 2023-04-29
RE2022-121 RE2022-124 Genome Attrib Selection UI and Related Table Refactor

dauglyon:
Apologies for another ginormous PR... I recommend reviewing the Table code commit-wise as there are a good amount changes due to the refactor. However, we already went over some of that I think, _and_ I have 100% test cov on Table, so tests could be a good place to start too.

- Adds selection to Genome Attribs
- adds helper buttons for (de)selecting matches
- adds checkboxes
- Table refactor for ease of use, better tests
- maintains 100% Test cov on Table :)

### DIFF
--- a/src/features/collections/MatchPane.tsx
+++ b/src/features/collections/MatchPane.tsx
@@ -5,11 +5,13 @@ import {
 } from '../../common/api/collectionsApi';
 import styles from './Collections.module.scss';
 import { useCallback, useEffect, useMemo, useState } from 'react';
-import { Select, SelectOption } from '../../common/components';
+import { Button, Select, SelectOption } from '../../common/components';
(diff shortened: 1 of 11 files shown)


src/features/collections/MatchPane.tsx
```diff
@@ -81,17 +84,25 @@ const ViewMatch = () => {
...
       <Button
         onClick={handleSelectAll}
-        disabled={
-          match?.state !== 'complete' ||
-          selectionSize + match.matches.length > 10000
+        disabled={match?.state !== 'complete' || matchTooLargeForSelection}
+        title={
+          matchTooLargeForSelection
+            ? 'Cannot select this match (too many items)'
+            : ''
```
dauglyon, line 105: added a title when the button is disabled for clarity

dakotablair (approved):

PR #77, opened 2023-05-05
RE2022-166 RE2022-167 Export

dauglyon:
- [Add basic export pane, api endpoints](https://github.com/kbase/ui-refresh-test/pull/77/commits/2c0417b657e3644827d9b60ba7332e1b46adb24b)
- [Use new+better narrative select in matching pane](https://github.com/kbase/ui-refresh-test/pull/77/commits/959e45a05109b44928cb4c505de6edc339eed690)
  - with the help of @dakotablair narrative selects are now way more usable

- [Display links to export results, and show export errors](https://github.com/kbase/ui-refresh-test/pull/77/commits/64504848912fcb68bc2e028c0c4d8f4a4955ab29)

- [Refactor selection to only create server selection when useSelectionI…](https://github.com/kbase/ui-refresh-test/pull/77/commits/468b6e6470fe73c21469255a9933e32bd8120d6e)
  - as noted by @MrCreosote, creating a selection for every change is overkill. This commit adds a hook and new serverside-selection-creation logic to only create an ID when a component using the `useSelectionId` hook is mounted

### DIFF
--- a/src/features/collections/ExportPane.tsx
+++ b/src/features/collections/ExportPane.tsx
@@ -2,10 +2,12 @@ import { useMemo, useState } from 'react';
 import {
   exportSelection,
   getSelectionTypes,
+  parseCollectionsError,
 } from '../../common/api/collectionsApi';
 import { getNarratives } from '../../common/api/searchApi';
 import { Select, Input, Button, SelectOption } from '../../common/components';
 import { useAppSelector } from '../../common/hooks';
+import { uriEncodeTemplateTag as encode } from '../../common/utils/stringUtils';
 import { useParamsForNarrativeDropdown } from './hooks';
 
 export const ExportPane = () => {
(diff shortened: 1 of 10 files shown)


src/features/collections/ExportPane.tsx
```diff
@@ -0,0 +1,124 @@
...
+    });
+  };
+
+  const exportError = parseCollectionsError(exportResult.error);
+  const exportErrorUnknown = exportResult.error && !exportError;
+  const exportErrorText = exportErrorUnknown
+    ? 'An unknown error occurred while saving to the narrative.'
+    : exportError
+    ? `${exportError.error.httpcode}: ${exportError.error.message}`
+    : undefined;
```
dakotablair, line 58: It looks like `exportError` is already known to be truthy, so this second ternary does not seem necessary. Edit: Oops that's so long as `exportErrorUnknown` is false, but the ternary covers when it is true. In that case we may want to change `undefined` to `''` as we discussed previously.

src/features/collections/MatchPane.tsx
```diff
@@ -144,16 +146,15 @@ const CreateMatch = ({ collectionId }: { collectionId: string }) => {
...
-  const narrativeQuery = listWorkspaceInfo.useQuery({});
-
-  const narrativeOptions = (narrativeQuery?.data?.[0] || []).map((ws) => ({
-    value: ws[0],
-    label: ws[1],
-    data: ws,
+  const narrativeSearchParams = useParamsForNarrativeDropdown(narrativeSearch);
+  const narrativeQuery = getNarratives.useQuery(narrativeSearchParams);
+  const narrativeOptions = (narrativeQuery?.data?.hits || []).map((hit) => ({
+    value: [hit.access_group, hit.obj_id, hit.version].join('/'),
```
dakotablair, line 154: Perhaps we should consider making a common function for formatting UPAs. I feel like I write it differently each time.

PR #80, opened 2023-05-11
  RE2022-173 Add microtrait API methods and useMicrotrait table hook stub

dauglyon:
Merging into `microtrait-all` (should I use `feature/microtrait`?) so I can keep microtrait things isolated from `main` for now. Also allows for a final review of `microtrait-all` before merging if we want.

see https://kbase-jira.atlassian.net/browse/RE2022-173 for ACs etc.


dakotablair (approved):
Looks like a good start!

PR #82, opened 2023-05-12
Update to React 18.

dakotablair:
I thought this might solve my problem with `<dialog />` but by the time I realized it wouldn't I was already committed, that is, the sunk cost fallacy had set in. So here it is!


dauglyon (approved):
ooh! LGTM

PR #86, opened 2023-05-26
RE2022-154 Adds Rank Parameter support to collections matching UI

dauglyon:
- @dakotablair we may want to discuss JSONSchema Libs. Here I added `ajv` as a proof of concept.  There are some (3 up-to-date, really) options though: https://json-schema.org/implementations.html#validator-javascript

### DIFF
--- a/src/features/collections/MatchPane.tsx
+++ b/src/features/collections/MatchPane.tsx
@@ -15,7 +15,7 @@ import { setUserSelection } from './collectionsSlice';
 import { store } from '../../app/store';
 import { useParamsForNarrativeDropdown } from './hooks';
 import { MatcherUserParams } from './MatcherUserParams';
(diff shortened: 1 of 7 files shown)


src/features/collections/MatchPane.tsx
```diff
@@ -14,6 +14,8 @@ import { useAppDispatch, useAppSelector, useBackoff } from '../../common/hooks';
 import { setUserSelection } from './collectionsSlice';
 import { store } from '../../app/store';
 import { useParamsForNarrativeDropdown } from './hooks';
+import { MatcherUserParams } from './MatcherUserParams';
+import { default as Ajv } from 'ajv';
```
dauglyon, line 18:
```suggestion
import Ajv from 'ajv';
```

PR #89, opened 2023-06-15
RE2022-198 Add homology matcher and improve error handling

dauglyon:


src/features/collections/MatcherUserParams.tsx
```diff
@@ -68,17 +87,44 @@ const GTDBLineageMatcherUserParameters = ({
...
+      <input
+        type="number"
+        min={max_dist?.minimum}
+        max={max_dist?.maximum}
+        step="0.01"
+        value={currentValue?.maximum_distance ?? max_dist?.default}
+        onChange={(e) => {
+          if (onChange)
+            onChange({
+              maximum_distance: !isNaN(parseFloat(e.currentTarget.value))
```
dakotablair, line 122: I think we need to use `Number.isNaN` here.

dakotablair (approved):
Looks good otherwise!

PR #92, opened 2023-06-28
URO-160: Create GHA, entrypoint, and hook it together to make deployment container image.

dauglyon:
In this PR is my idea for image creation, based off what @dakotablair and I discussed. 

`$STATIC_ENVIRONMENT` needs to be set when the container is run, which will determine which environment build the images serves.

- [x] Statically build Europa for each environment.
- [ ] Push image to `ghcr`.
- [x] On container startup, rewrite `nginx` config to serve correct build based on an environment variable.
- [x] Can be deployed in rancher.


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM bitnami/nginx:latest
+USER root
+
+# Copy built static files for all enviroments to image
+COPY ./deploy /deploy/
+
+# Copy nginx config template to image
+COPY ./scripts/nginx.conf.tmpl /nginx.conf.tmpl
+
+COPY ./scripts/entrypoint.sh /entrypoint.sh
```
dauglyon, line 10:
```suggestion
COPY ./scripts/docker_entrypoint.sh /entrypoint.sh
```

src/common/components/PlaceholderFactory.tsx
```diff
@@ -12,7 +12,7 @@ export const PlaceholderFactory = (name: string) => {
   const Placeholder: FC<PlaceholderInterface> = (props) => {
     const invalidEnvironment =
```
bio-boris, line 13: // This placeholder factory makes components that can be used as stubs for
// development. There should be no Placeholders in production, so an error is
// thrown in that case.
There are 2 production environments, appdev, and production, if that matters

scripts/build_deploy.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+
+declare -a enviroments=(
+# "<name>         <domain>                <legacy-domain>"
+  "ci-europa      ci-europa.kbase.us      legacy.ci-europa.kbase.us"
+  "narrative-dev  narrative-dev.kbase.us  legacy.narrative-dev.kbase.us"
```
bio-boris, line 7: This appears to be missing
* legacy.ci.kbase.us
* legacy.next.kbase.us
* legacy.staging.kbase.us

bio-boris (commented):
Very cool!

scripts/build_deploy.sh
dauglyon: None of those environments exist at this time

dauglyon: They'll need to, eventually but they're not there yet...

dauglyon:
Closing, changes taken up in https://github.com/kbase/ui/pull/95

PR #101, opened 2023-08-08
RE2022-225 RE2022-200 Phase 0 Collections Detail page (and everything else)

dauglyon:
See https://kbase-jira.atlassian.net/browse/RE2022-225

Also fixes https://kbase-jira.atlassian.net/browse/RE2022-200

Screenshot 2023-08-09 at 10 42 50 AM

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -16,7 +16,7 @@ import Routes from './Routes';
 import LeftNavBar from '../features/layout/LeftNavBar';
 import TopBar from '../features/layout/TopBar';
 import ErrorPage from '../features/layout/ErrorPage';
-import { ModalProvider } from '../features/layout/Modal';
+import { ModalDialog } from '../features/layout/Modal';
 
 const useInitApp = () => {
   const dispatch = useAppDispatch();
@@ -54,13 +54,8 @@ export default function App() {
           </div>
           <div className={classes.page_content}>
             <ErrorBoundary FallbackComponent={ErrorPage}>
-              {isLoading ? (
-                'Loading...'
-              ) : (
-                <ModalProvider>
-                  <Routes />
-                </ModalProvider>
-              )}
+              {isLoading ? 'Loading...' : <Routes />}
+              <ModalDialog />
             </ErrorBoundary>
           </div>
         </div>
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -16,6 +16,7 @@ import Routes from './Routes';
 import LeftNavBar from '../features/layout/LeftNavBar';
 import TopBar from '../features/layout/TopBar';
 import ErrorPage from '../features/layout/ErrorPage';
+import { Modal } from '../features/layout/Modal';
 
 const useInitApp = () => {
   const dispatch = useAppDispatch();
@@ -53,7 +54,14 @@ export default function App() {
           </div>
           <div className={classes.page_content}>
             <ErrorBoundary FallbackComponent={ErrorPage}>
-              {isLoading ? 'Loading...' : <Routes />}
+              {isLoading ? (
+                'Loading...'
+              ) : (
+                <>
+                  <Routes />
+                  <Modal />
+                </>
+              )}
             </ErrorBoundary>
           </div>
         </div>
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -16,7 +16,7 @@ import Routes from './Routes';
 import LeftNavBar from '../features/layout/LeftNavBar';
 import TopBar from '../features/layout/TopBar';
 import ErrorPage from '../features/layout/ErrorPage';
-import { Modal } from '../features/layout/Modal';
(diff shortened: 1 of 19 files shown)


dauglyon:
Not sure what is happening with the build failing... Locally it compiles with warnings. Maybe the warnings are causing the fail?

src/features/layout/Modal.module.scss
```diff
@@ -0,0 +1,45 @@
...
+  $padding: 1em;
+  $border: 1px solid use-color("base-light");
+  border: $border;
+  display: flex;
+  flex-flow: column nowrap;
+  overflow: visible;
+  padding: $padding;
+
+  .header {
+    align-items: start;
```
dakotablair, line 13: I think this is what the build is complaining about. It wants you to use `flex-start` instead.

src/features/layout/Modal.module.scss
```diff
@@ -0,0 +1,45 @@
...
+      font-weight: normal;
+    }
+  }
+
+  .footer {
+    border-top: $border;
+    display: flex;
+    flex-direction: row;
+    gap: 0.333em;
+    justify-content: end;
```
dakotablair, line 31: See also the other comment, I think the linter wants you to use `flex-end` here.

src/features/layout/Modal.module.scss
dauglyon: funky. I wonder if there is a linter setting given the warning is for browsers we don't support.

src/common/components/Table.module.scss
codytodonnell, line 1: Generally I prefer to use `rem` instead of `em` to keep sizing predictable. I know a `rem` value in one place will be the same as a `rem` value in another.

src/common/components/Button.tsx
```diff
@@ -8,6 +8,7 @@ type Color = typeof colors[number];
 export interface CommonButtonProps {
   variant?: 'contained' | 'outlined' | 'text';
   color?: Color;
+  textColor?: Color;
```
dakotablair, line 11: I have a suspicion that we may regret not sticking to `backgroundColor` and `color` here.

src/app/App.tsx
```diff
@@ -53,7 +54,13 @@ export default function App() {
           </div>
           <div className={classes.page_content}>
             <ErrorBoundary FallbackComponent={ErrorPage}>
-              {isLoading ? 'Loading...' : <Routes />}
+              {isLoading ? (
+                'Loading...'
```
dakotablair, line 58: Consider `<>Loading...</>`.

src/features/collections/CollectionDetail.tsx
```diff
@@ -96,13 +191,10 @@ export const CollectionDetail = () => {
               dataProduct={currDataProduct}
               collection_id={collection.id}
             />
-          ) : (
-            <div className={styles['data_product_placeholder']}>
-              <span>Select a Data Product</span>
-            </div>
-          )}
+          ) : null}
```
dakotablair, line 194: Consider `<></>`.

src/features/layout/Modal.module.scss
dakotablair: Turns out it is not the linter (since _the linter is not running_!) so I am not totally sure where to tell it to chill. For the purposes of this PR I'm ok with this being the only problem. See #102 for other exciting fixes like: actually run the linter.

dakotablair (approved):
Looks good! I think we should merge it and fix any problems that arise later.

src/common/components/Button.tsx
codytodonnell: I'm thinking this may be temporary until we can go in and set up text colors explicitly for the different button types (primary, secondary, etc.). We could consider keeping it around for special cases or overriding with css in special cases. I'm not sure which is better.

src/common/components/Table.module.scss
dauglyon: do we want to discuss this as a dev guideline? I tend to use em but that's mostly because I learned before rem was in common use.

src/features/collections/DataProduct.tsx
```diff
@@ -9,15 +9,16 @@ export const DataProduct: FC<{
...
+    case 'taxa_count':
+      return <TaxaCount {...{ collection_id }} />;
+    case 'genome_attribs':
+      return <GenomeAttribs {...{ collection_id }} />;
+    case 'microtrait':
+      return <Microtrait {...{ collection_id }} />;
+    case 'samples':
+      return <SampleAttribs {...{ collection_id }} />;
+    default:
+      return <>No view implemented for DataProduct '{dataProduct.product}'</>;
```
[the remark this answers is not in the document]
dauglyon: But, there is a default case.

src/common/components/Table.module.scss
codytodonnell: Sure let's bring it up in the next UI meeting

PR #106, opened 2023-08-17
Add simple/placeholder Loader component

dauglyon:
Adds a really simple loader component. This will need more presentation work in the future, but it will be nice to have a common component for this.

### DIFF
--- a/src/common/components/Loader.tsx
+++ b/src/common/components/Loader.tsx
@@ -1,0 +1,29 @@
+import { faSpinner } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+/**
+ * Component for rendering loading states
+ */
+export const Loader = (props: {
+  /**state of loader, defaults to true */
+  loading?: boolean;
+  /**style of loader to render when loading==true */
+  type?: 'text' | 'spinner';
+  /**Loader children are rendered when loading==false */
+  children?: React.ReactNode | React.ReactNode[];
+  /** Overrides the loader style to render a custom loader*/
+  render?: React.ReactNode | React.ReactNode[];
+}) => {
+  if (props.loading ?? true) {
+    if (props.render !== undefined) return <>{props.render}</>;
+    switch (props.type) {
+      case 'text':
+        return <>Loading...</>;
+      case 'spinner':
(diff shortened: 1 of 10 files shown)


src/common/components/Loader.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const Loader = (props: {
+  /**state of loader, defaults to true */
+  loading?: boolean;
+  /**style of loader to render when loading==true */
+  type?: 'text' | 'spinner';
+  /**Loader children are rendered when loading==false */
+  children?: React.ReactNode | React.ReactNode[];
+  /** Overrides the loader style to render a custom loader*/
+  render?: React.ReactNode | React.ReactNode[];
+}) => {
```
dakotablair, line 16: I've got a stylistic question for you, but do not let it distract you too much.

Should we put more effort into a unified code style for functional components?

In abstract there is not much difference I can see between this version and one more like what I probably would have written:
```js
import { FC } from 'react';

interface LoaderProps {
  /**Loader children are rendered when loading==false */
  children?: React.ReactNode | React.ReactNode[];
  /**state of loader, defaults to true */
  loading?: boolean;
  /** Overrides the loader style to render a custom loader*/
  render?: React.ReactNode | React.ReactNode[];
  /**style of loader to render when loading==true */
  type?: 'text' | 'spinner';
}

export const Loader: FC<LoaderProps> = ({
  children,
  loading,
  render,
  type,
}) => {
  console.log({ children, loading, render, type });
  // etc...
};
```

In this case, I think there are not many differences between these approaches. The one that comes to my mind is that it is easier to export `LoaderProps` in case you need to reuse it (which is not _yet_ a concern for this component).

What do you think?

src/common/components/Loader.tsx
```diff
@@ -0,0 +1,29 @@
...
+  if (props.loading ?? true) {
+    if (props.render !== undefined) return <>{props.render}</>;
+    switch (props.type) {
+      case 'text':
+        return <>Loading...</>;
+      case 'spinner':
```
dakotablair, line 22: 😱 Fallthrough! AAH!
```python
"""cries"""
```
(just kidding, of course)

dakotablair (approved):
This is just better. What a great PR!

PR #107, opened 2023-09-11
URO-163, URO-165, URO-166: API calls for Copy, Rename and Restore components

dakotablair:
- [ ] URO-163: Narratives can be copied to a new narrative with the specified name using the Narrative Control Menu.
  - [ ] If the action fails then this is shown to the user using a toast notification. Have distinct messages for when:
    - [ ] The specified name is not valid (although, this should probably be enforced through UI validation).
    - [ ] The app cannot confirm action success.
- [ ] URO-165: Narratives can be renamed to the specified name using the Narrative Control Menu.
  - [ ] If the action fails then this is shown to the user using a toast notification. Have distinct messages for when:
    - [ ] The specified name is not valid (although, this should probably be enforced through UI validation).
    - [ ] The app cannot confirm action success.
- [ ] URO-166: Narratives can be restored to the specified version using the Narrative Control Menu.
  - [ ] If the action fails then this is shown to the user using a toast notification. Have distinct messages for when:
    - [ ] The app cannot confirm action success.


src/common/api/orgsApi.ts
```diff
@@ -23,11 +23,13 @@ export interface OrgMemberInfo {
 export interface OrgsParams {
   getNarrativeOrgs: number;
   getUserOrgs: void;
+  linkNarrative: { orgId: string; wsId: number };
 }
 
 export interface OrgsResults {
   getNarrativeOrgs: OrgInfo[];
   getUserOrgs: OrgMemberInfo[];
+  linkNarrative: unknown;
```
dauglyon, line 32: What's this unknown? Is there no response or is it not well defined?

src/common/api/serviceWizardApi.ts
```diff
@@ -1,34 +1,40 @@
+/* serviceWizardApi */
 import { baseApi } from './index';
 import { setConsumedService } from './utils/kbaseBaseQuery';
 import { jsonRpcService } from './utils/serviceHelpers';
 
 const serviceWizard = jsonRpcService({ url: 'services/service_wizard' });
 
+const dynamicService = jsonRpcService;
+
+/* Use this for dynamic services to ensure serviceWizardApi is set. */
```
dauglyon, line 10: Nice!

src/common/components/Input.tsx
```diff
@@ -60,9 +61,10 @@ export const Input = forwardRef<HTMLInputElement, InputInterface>(
           {...rest}
           className={classes.input}
           id={idForLabel}
-          type={'text'}
-          onFocus={handleFocus}
+          maxLength={maxLength}
           onBlur={handleBlur}
+          onFocus={handleFocus}
+          type={'text'}
```
dauglyon, line 67: If these should be "defaults" and overridable by props on the Input element, they should be above the {...rest} call, so that they get overridden

dauglyon: IIRC

src/features/navigator/NarrativeControl/Copy.tsx
```diff
@@ -21,23 +24,50 @@ export interface CopyProps extends ControlProps {
...
+      dispatch(setLoading(false));
+    } catch (err) {
+      if (!isKBaseBaseQueryError(err)) {
+        console.error({ err }); // eslint-disable-line no-console
+        toast(ErrorMessage({ err }));
+        return;
+      }
+      toast(ErrorMessage({ err: parseError(err) }));
+      dispatch(setLoading(false));
+      return;
```
dauglyon, line 66: This seems like something we may do a lot, perhaps worth pulling out into a util, I'm not sure. What do you think?

src/features/navigator/NarrativeControl/Restore.tsx
```diff
@@ -1,29 +1,71 @@
...
+    } catch (err) {
+      if (!isKBaseBaseQueryError(err)) {
+        console.error({ err }); // eslint-disable-line no-console
+        toast(ErrorMessage({ err }));
+        return;
+      }
+      toast(ErrorMessage({ err: parseError(err) }));
+      dispatch(setLoading(false));
+      return;
+    }
```
dauglyon, line 51: Potential duplicate of the chunk I mentioned

dauglyon (approved):
Some minor comments, but approved

src/features/navigator/NarrativeControl/Copy.tsx
dakotablair: To do that we would need to pass `dispatch` as a parameter, which is somewhat standard, but still a little less readable to developers not used to a more functional style. We'll keep an eye on it.

src/common/components/Input.tsx
dakotablair: Yes, I think you're right and at least `maxLength` should go above `{...rest}` I'm not sure about the others, it depends on how much of our will we want to impose on users of this component "by default" since if one needed an override they could fix it here too.

PR #109, opened 2023-09-20
URO-82, URO-168 Add share narrative function.

dakotablair:
This PR adds the API calls for updating sharing in the narrative control menu. This is also the final piece of functionality for the Navigator as a whole. That is, no major new functionality remains for Navigator itself. In particular, the narrative navigator is now feature complete.

- [x] URO-82: Create Navigator Component
  - [x] Aligns roughly to mockup URO-78.
  - [x] Implement user behaviors existing in previous Navigator
  - [x] Manages state for searches and filters
  - [x] Manages quantity of search results displayed
  - [x] Manages state for the control menu including
    - [x] delete
    - [x] copy
    - [x] link (to an organization)
    - [x] rename
    - [x] restore
    - [x] share
  - [x] Handles narrative metadata changes at least to the standard of the previous Navigator
- [x] URO-168: Add API calls to the Share Narrative action in the control menu.
  - [x] Narratives can be shared (Narrative permissions of a specified level can be assigned to a specified user) using the Narrative Control Menu. Specifically narrative admins may:
    - [x] Add a new user with selected permissions
    - [x] Remove a user from shares
    - [x] Update a user's permission level
    - [x] Make a narrative globally public or private
  - [x] If the action fails then this is shown to the user using a toast notification. Have distinct messages for when:
    - [x] The app cannot confirm action success.

### DIFF
--- a/src/features/navigator/NarrativeControl/Share.tsx
+++ b/src/features/navigator/NarrativeControl/Share.tsx
@@ -1,27 +1,32 @@
 /* NarrativeControl/Share */
 import { FontAwesomeIcon as FAIcon } from '@fortawesome/react-fontawesome';
 import { faLock, faUnlock } from '@fortawesome/free-solid-svg-icons';
+import toast from 'react-hot-toast';
 import { FC, ReactElement, useEffect, useId, useMemo, useState } from 'react';
 import { authToken, authUsername } from '../../auth/authSlice';
 import { searchUsers } from '../../../common/api/authService';
-import { getwsPermissions } from '../../../common/api/workspaceApi';
+import { isKBaseBaseQueryError } from '../../../common/api/utils/common';
+import { parseError } from '../../../common/api/utils/parseError';
+import {
+  getwsPermissions,
+  setwsGlobalPermissions,
+  setwsUsersPermissions,
+} from '../../../common/api/workspaceApi';
 import { Button, Select, SelectOption } from '../../../common/components';
 import { useAppDispatch, useAppSelector } from '../../../common/hooks';
 import { NarrativeDoc } from '../../../common/types/NarrativeDoc';
 import {
   removeShare,
+  setLoading,
   setShares,
   setUserPermission,
   shares,
+  synchronized,
   updateUsers,
   users,
 } from '../navigatorSlice';
-import {
-  isUserPermission,
-  permissions,
-  TODOAddLoadingState,
-  UserPermission,
-} from '../common';
+import { isUserPermission, permissions, UserPermission } from '../common';
+import { ErrorMessage } from './common';
 import classes from './NarrativeControl.module.scss';
 
 const permissionValues: Record<UserPermission, SelectOption> = {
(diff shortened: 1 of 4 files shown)


src/features/navigator/NarrativeControl/Share.tsx
```diff
@@ -33,17 +38,24 @@ const permissionValues: Record<UserPermission, SelectOption> = {
...
+  const syncd = useAppSelector(synchronized);
+
+  /* derived values */
   const permissionOptions = [
     permissionValues.r,
     permissionValues.w,
     permissionValues.a,
   ];
+
+  /* <UserPermissionSelection /> component */
```
dauglyon, line 55: what is this comment? Is this a planned feature? Or something we chose not to implement?

src/features/navigator/NarrativeControl/Share.tsx
```diff
@@ -223,44 +341,74 @@ export const Share: FC<{
...
     }
   }, [dispatch, wsPermsQuery.currentData, wsPermsQuery.isSuccess]);
-  const shareSelectId = useId();
+
+  /* derived values */
   const narrativeIsPublic = usernameShares['*'] === 'r';
   const userPermission: UserPermission =
     (usernameAuthed && usernameShares[usernameAuthed]) || 'n';
-  const userIsAdmin = usernameAuthed && userPermission === 'a';
+  const userIsAdmin = Boolean(usernameAuthed) && userPermission === 'a';
```
dauglyon, line 349: not sure Boolean is needed, but no need to remove it

src/features/navigator/NarrativeControl/Share.tsx
```diff
@@ -223,44 +341,74 @@ export const Share: FC<{
...
+  });
+
+  /* add permission callback */
+  // See <SelectUser /> component
+
+  /* set permission callback */
+  // See <UserPermissionControl /> component
+
+  /* remove permission callback */
+  // See <UserPermissionControl /> component
```
dauglyon, line 363: I'm not sure I get why these comments are here as things like `setUserPermission` are already listed as imports

dauglyon (approved):
minor comments, but everything LGTM

src/features/navigator/NarrativeControl/Share.tsx
dakotablair: The `usernameAuthed` is actually a string. I think the name is a bit confusing, so I can understand this comment. Then `userIsAdmin` is passed as a prop to `PublicToggle` which specifies a boolean.

src/features/navigator/NarrativeControl/Share.tsx
dakotablair: It's just supposed to demarcate the "getting ready to render" part of the component from the actual render biz. There are a few other comments with a similar purpose lying around because these components are somewhat complex.

src/features/navigator/NarrativeControl/Share.tsx
dakotablair: I have been organizing these control menu components like this:
```js
/* hooks */
/* derived values */
/* callback */
/* render */
```

except Share has 4 different callbacks and it made the most sense to me to put them in these other components, so it's just a reminder (mainly for me) to know where to look for the callback corresponding to the listed behavior. I was also thinking that this module is getting a little long and could probably serve to be split, but never actually did that since it ended up less than 500 lines for now.

PR #110, opened 2023-09-27
URO-176: Narrative loading and linking in Europa

dakotablair:
- [ ] URO-176:
  - [x] Links to narratives in Europa should be environment aware and link to the appropriate narrative.
  - [ ] Narrative containers can take some time before they are ready, and users should not be left stranded with a page that says their narrative cannot be loaded, if at all possible.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -38,11 +38,14 @@ const useInitApp = () => {
   return { isLoading: !initialized };
 };
 
+/*
+    <Router basename={process.env.PUBLIC_URL}>
+*/
 export default function App() {
   const { isLoading } = useInitApp();
 
   return (
-    <Router basename={process.env.PUBLIC_URL}>
+    <Router>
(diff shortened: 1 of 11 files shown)


src/app/App.tsx
```diff
@@ -38,11 +38,14 @@ const useInitApp = () => {
...
 
+/*
+    <Router basename={process.env.PUBLIC_URL}>
+*/
 export default function App() {
   const { isLoading } = useInitApp();
 
   return (
-    <Router basename={process.env.PUBLIC_URL}>
+    <Router>
```
dakotablair, line 48: This kind of change is actually opposed to a minor goal of Europa, that is, to not make the assumption that Europa has control over the absolute routes. Therefore, in the future, if for some reason (perhaps when it is replaced with something new), it can be served from the same host as the new thing, making cookie management simpler.

PR #115, opened 2023-10-09
URO-178: Add feeds notifcation count, feeds api

dauglyon:
- [x] URO-178: Number of unread feeds should appear on the feeds icon.
  - [x] The number of unread feeds should appear on the Feeds icon.

### DIFF
--- a/src/features/layout/LeftNavBar.module.scss
+++ b/src/features/layout/LeftNavBar.module.scss
@@ -17,6 +17,7 @@ ul.nav_list {
     flex-flow: column nowrap;
     margin: 3px 0;
     padding: 3px;
+    position: relative;
     text-align: center;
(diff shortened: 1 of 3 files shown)


src/features/layout/LeftNavBar.module.scss
```diff
@@ -42,5 +43,17 @@ ul.nav_list {
     .nav_desc {
       font-size: 14px;
     }
+
+    .nav_notifs {
+      background-color: use-color("error");
```
dakotablair, line 48: I know that the notification is red today, but perhaps this is another good place to introduce our semantic coloring? That is, unseen notifications are not errors, more like info?

dakotablair (approved):

PR #116, opened 2023-10-11
URO-179: Add /narrative/ fallback case.

dakotablair:
- [x] URO-179: Add /narrative/ fallback case.
  - [x] When `https://DOMAIN/legacy/narrativemanager/new` is loaded then `legacy.DOMAIN` redirects to a URL like `https://DOMAIN/fallback/narrative/69420` at which point Europa should redirect to the correct narrative, that is, `https://DOMAIN/narrative/69420` at the top level.


src/app/App.tsx
```diff
@@ -38,30 +39,35 @@ const useInitApp = () => {
   return { isLoading: !initialized };
 };
 
-export default function App() {
+const App: FC = () => {
   const { isLoading } = useInitApp();
-
+  const location = useLocation();
+  const fallbackClasses =
```
dauglyon, line 45: I don't think this is the right way to do this, we should probably add a util hook that can be called from within a page component to disable the chrome

dauglyon: But I agree fallback should have the chrome disabled

src/features/legacy/Fallback.tsx
```diff
@@ -1,33 +1,38 @@
...
  * 404s from the legacy site are redirected from legacy.DOMAIN/[some/path/here] to DOMAIN/fallback/[some/path/here]
  * this component handles these fallback redirects for pages such as '/narratives'
  */
 export const Fallback = () => {
-  const location = useParams();
   const navigate = useNavigate();
+  const params = useParams();
 
-  const fallbackPath = location['*']?.toLowerCase();
+  const fallbackPath = params['*']?.toLowerCase();
```
dauglyon, line 14: Ooops, nice catch

src/features/legacy/Fallback.tsx
```diff
@@ -1,33 +1,38 @@
...
       // Not in top window, redirect top window to current location
       window.top.location = window.location;
+      return;
+    }
+
+    // We are in the top window
+    // redirect appropriately based on the fallbackPath, otherwise set notFound to true
+    if (fallbackPath === 'narratives') {
+      navigate('/narratives');
+    } else if (Object.hasOwn(params, 'wsId')) {
```
dauglyon, line 29: Maybe we can use route matching here, instead of relying on string equality or params? Let me find the doc link

dauglyon (approved):

PR #117, opened 2023-10-11
URO-177: Align favicon to production KBase favicon.

dakotablair:
- [x] URO-177:
  - [x] The favicon in Europa is the same as the favicon on [kbase.us](http://kbase.us/).


public/manifest.json
```diff
@@ -1,21 +1,11 @@
 {
-  "short_name": "React App",
-  "name": "Create React App Sample",
+  "short_name": "KBase",
```
dauglyon, line 2: KBase/ui?

src/app/App.test.tsx
```diff
@@ -36,7 +39,9 @@ describe('The main UI for Europa...', () => {
   test('contains a working link to the Navigator', async () => {
     const { container } = render(
       <Provider store={createTestStore()}>
-        <App />
```
dauglyon, line 39: I think there's more in this PR than you intended

dakotablair: Now that the fallback code has been merged, I went ahead and rebased this branch to take it out of this PR.

public/manifest.json
dakotablair: Updated `name` to be `kbase/ui` and removed `short_name` since `name` is less than 12 characters.

dauglyon (approved):

PR #118, opened 2023-10-11
nf opt2, parameterize in Routes.tsx, eliminate useEffect

dauglyon:


dakotablair:
I forgot to approve this before I merged it, but it looked good except for the one thing which I fixed. This also addresses my concern about the fallback routes and what to do with them living in separate places, so that's awesome! 🎉

PR #120, opened 2023-10-14
Navigator UI Tweaks

codytodonnell:
Some smalls UI changes to the navigator UI as mentioned in my comments in the navigator feedback doc:

- Add padding to narrative metadata
  - Improves readability and aesthetic of the metadata tables
- Add space to the left of narrative actions menu
  - These two elements were right up against each other. This is just a small visual improvement.
- Fix narrative title alignment and icon position
  - This fixes a slight misalignment with the narrative title and the icons in that component
  - This also moves the new-window icon to the right of the title. It's more common to see this icon at the end of a link and it helps improve the readability of the link content.
- Fix author name overflow and add better responsiveness to metadata columns
  - This fixes an issue with some metadata columns overflowing outside of the gray box.
  - This also changes the way the metadata column widths work, opting to have them expand to fill the whole width and wrap at smaller widths. I thought this was a visual improvement.
  - This also adds a fixed width to the bolded column in LabelValueTables so that values never get too far from their label and so that tables stacked on top of each other have values all aligned.

### DIFF
--- a/src/common/components/LabelValueTable.module.scss
+++ b/src/common/components/LabelValueTable.module.scss
@@ -5,14 +5,13 @@
 
 .label-value-table td {
   padding: 0;
-  white-space: nowrap;
 }
 
 .label-value-table th {
   font-weight: bold;
   text-align: left;
   vertical-align: top;
-  white-space: nowrap;
+  width: 10rem;
 }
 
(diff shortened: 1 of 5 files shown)


dakotablair (approved):
Looks great!

PR #128, opened 2023-11-28
URO 95: Configure Google Analytics 4

dakotablair:
- [x] Add G4A and GTag tracking, implement and use appropriate hooks for both, integrating with react-router
- [x] MUST support legacy routes and new routes
- [x] Coordinate with credit-engine team about user-id tracking.
- [x] Request review from @briehl

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,6 +1,15 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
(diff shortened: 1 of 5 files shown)


briehl (approved):
I can't speak to the rest of the PR, but the last 4 commits around GA4 look good to me.

Thanks @dakotablair !

PR #133, opened 2023-12-08
RE2022-278 RE2022-280 add basic histogram and scatter

dauglyon:
Also solves the following bug punchlist issues:
- In ENIGMA genome attribs there’s a static spinner - I think this indicates a missing graph
Investigate 2d histogram failure states
Better loading state (same size as graph)
- In ENIGMA genome attribs there’s no data points in the completeness / contamination chart.
Data loading issue, keys are different between gtdb and enigma, should be solved before deploy
Need to special case for GTDB


dauglyon:
The scatter plot is disappearing on zoom; it needs to be fixed before merging. Also, check the issue for scroll formatting.

codytodonnell (approved):
Looks fantastic!

I am having an issue installing the new dependencies locally. This appears to be specific to my local environment and is caused by the `canvas` dev dependency. See relevant thread: https://github.com/Automattic/node-canvas/issues/1773#issuecomment-1385216460

dakotablair (approved):
We may be able to fix the `canvas` build problems with the suggestions in `node-canvas` [issue #2353](https://github.com/Automattic/node-canvas/issues/2353).

PR #135, opened 2023-12-19
DEVOPSs1412 :  Europa on Next

jsfillman:
After exhaustive troubleshooting, I found the root cause behind the `404` errors on the `next` cluster:

``The next environment wasn't defined in Europa's env configs!``

I've updated the relevant files to include, and thus build code for, a `next` env, which now shows up under `/deploy/next`.
I've gone ahead and tested this in the next cluster using image ``ghcr.io/kbase/ui:br-DEVOPS-1412-next``.

### DIFF
--- a/config.json
+++ b/config.json
@@ -24,6 +24,12 @@
       "public_url": "/"
     },
 
+    "next": {
+      "domain": "next.kbase.us",
+      "legacy": "legacy.next.kbase.us",
+      "public_url": "/"
+    },
+
     "production": {
       "domain": "narrative.kbase.us",
       "legacy": "legacy.narrative.kbase.us",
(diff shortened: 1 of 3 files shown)


config.json
```diff
@@ -24,6 +24,12 @@
       "public_url": "/"
     },
 
+    "next": {
+      "domain": "next.kbase.us",
+      "legacy": "legacy.next.kbase.us",
+      "public_url": "/"
+    },
+
```
dakotablair, line 32: This is the only place that was required to be updated to prevent the 404s ran into. It is the situation "using a deployed build with an undefined environment" which usually does not happen during development since the defaults from the `.env` file are used. We might want to update `scripts/build_deploy.sh` to use defaults from `.env`.

dakotablair (approved):
Looks good to me!

PR #136, opened 2024-02-01
Update Collection filter panel layout

codytodonnell:
- Moves Collection filter menu into a panel in the main content area
- Makes filter panel expand to fill height of window and lets users scroll to see additional filters
- Right side of content area is pushed to the right when panel is open
- Elements in data products have consistent panels wrapped around them
- Adds padding around pagination components and table toolbars
- Adds row count text above tables and heatmaps

- [x] Add rowCount to useMicrotrait hook and genome attributes table hook


dauglyon (approved):
Approved pending conflict resolution and fixing the row count

PR #143, opened 2024-02-14
Update table UI

codytodonnell:
Update the style of the table headers and use the display name for their text instead of the field name. Also use the column type to change the alignment for number columns.

Also includes some style changes to the sidebar.


dauglyon (approved):
Lgtm!

PR #144, opened 2024-02-14
Fix GAttribs table pagination

dauglyon:


codytodonnell (approved):
Working on my end! I don't entirely understand what line 279 is doing but this definitely solved the issue.

dauglyon:
It's saying whether or not to skip the query on that run of the hook. So the query was remaining uncalled because !![variable] was always true

dauglyon:
Which makes the pagination always default to length zero because a count was never returned from the server (because it was never called)

PR #146, opened 2024-02-26
PUBLIC-2722 Fix Google Re-login bug

dauglyon:
1. Before this bugfix, Europa did its best to isolate its own TLD cookie from legacy.* domains. However, this was causing a mismatch in the auth state between Europa and Legacy.
2. Cookie expiration in Europa was running AFTER cookie checks in legacy
3. Legacy flow fails for Google if an invalid cookie is present AT ALL during the login/continue flow endpoint, but it wasn't deleted before the flow, as Europa-cookie-clearing was losing the race condition when the intermediate login page (account choice) wasn't present. It was also losing the race condition on the login page itself and redirecting there when cookies were invalid (meaning cookies were never being deleted)
4. Europa no longer "hesitates" to delete bad cookies until after legacy load (resolving the race condition), and makes sure the domain is always specified when clearing the cookie (although this turned out not to be the issue, it is a latent bug)
5. Cookies are also now properly (completely) deleted on logout!

See comments for specific code change locations

### DIFF
--- a/src/features/auth/hooks.ts
+++ b/src/features/auth/hooks.ts
@@ -34,13 +34,19 @@ export const useTokenCookie = (
   const dispatch = useAppDispatch();
 
   // Pull token from main cookie. If it exists, and differs from state, try it for auth.
-  const [cookieToken, setCookieToken, clearCookieToken] = useCookie(cookieName);
+  const [cookieToken, setCookieToken, clearCookieToken] = useCookie(
+    cookieName,
+    process.env.NODE_ENV === 'development'
+      ? {}
+      : { domain: `.${process.env.REACT_APP_KBASE_DOMAIN}` }
+  );
+
   const { isSuccess, isFetching, isUninitialized } =
     useTryAuthFromToken(cookieToken);
 
   // Controls for backupCookie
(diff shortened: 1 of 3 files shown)


src/features/auth/hooks.ts
```diff
@@ -34,37 +34,65 @@ export const useTokenCookie = (
...
   // Pull token, expiration, and init info from auth state
   const token = useAppSelector(authToken);
   const expires = useAppSelector(({ auth }) => auth.tokenInfo?.expires);
-  const initialized = useAppSelector(authInitialized);
+  const appAuthInitialized = useAppSelector(authInitialized);
 
   // Initializes auth for states where useTryAuthFromToken does not set auth
   useEffect(() => {
-    if (isFetching || initialized) return;
+    // If the cookieToken is present but it failed checks and wont be overwritten by a token in state, clear
```
dauglyon, line 58: this is the new condition to clear the cookieToken immediately after the token check call, which always happens before the legacy iframe is loaded (instead of waiting for the iframe to postMessage its login state). This is safe, as the cookie now uses the same domain for Legacy and Europa.

codytodonnell (approved):

PR #147, opened 2024-02-26
Update README.md

dauglyon:


dakotablair:
While we are at it, let's add the test badge: build and push status

PR #149, opened 2024-02-27
Bugfixes to overview page.

codytodonnell:
Fix taxa chart menu overflow. Make species the default option in taxa chart. Adjust panel spacing.


dauglyon (approved):
lgtm!

PR #159, opened 2024-02-28
fix exportSelection url

dauglyon:


dakotablair (approved):
We are forced to honor the backend endpoint URLs.

PR #132, opened 2023-12-04
RE2022-241: Microtrait update

dakotablair:
- [x] [RE2022-241](https://kbase-jira.atlassian.net/browse/RE2022-241): HeatMap component refactor
  - [x] tooltips
    - [x] cursor following tooltip which locks on click
    - [x] content selectable for clipboard management
  - [x] constrain axis ranges to data
    - [x] If a user navigates outside the range of the data, the view should revert to showing the data.
  - [x] labels render vertically
  - [x] some things we get for free with plotly.js:
    - [x] quick rendering of ~30k cells
    - [x] zoom, scroll
    - [x] labels adjust to available space
    - [x] column labels at top
    - [x] update width on window resize

[RE2022-241]: https://kbase-jira.atlassian.net/browse/RE2022-241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #164, opened 2024-03-01
plotly.js HeatMap

dakotablair:
HeatMap hot off the presses, perhaps a little half-baked, ready for Sean's preliminary review on 2024-03-01.


dakotablair:
See #132.

PR #165, opened 2024-03-05
Add onChange debounced submits for filter text fields

dauglyon:
fixes bug:
> Improve submit conditions on text fields in filter controls (enter should filter, maybe debounce onChange along with onBlur)


dakotablair:
This also appears to address URO-141 and may also be usable as is to fix some issues with heatmap tooltips.

dakotablair (approved):

PR #167, opened 2024-03-11
Add troubleshooting notes to README for canvas dependency issue

codytodonnell:


dauglyon (approved):
Add info about the specific context of this error and the cause in your case

PR #169, opened 2024-03-11
Fix Export Form

dauglyon:
fixes three punchlist bugs:

- `Don’t show read-only narratives in the export select box`
- `Export form needs improved error states`
- `In export GUI, Data Object Name -> New [Genome | Assembly] Set Object Name`

also fixes input blur/focus handler forwarding and disables the export button in the select modal when there is no selection.

#### Success
Screenshot 2024-03-11 at 3 20 52 PM

#### Validation Error
Screenshot 2024-03-11 at 4 06 14 PM

#### Server Error (forced this error because I couldn't figure out a valid form state that lead to a server error)
Screenshot 2024-03-11 at 3 29 49 PM

### DIFF
--- a/src/common/api/collectionsApi.ts
+++ b/src/common/api/collectionsApi.ts
@@ -632,11 +632,17 @@ export const collectionsApi = baseApi.injectEndpoints({
       CollectionsResults['exportSelection'],
       CollectionsParams['exportSelection']
     >({
-      query: ({ selection_id, workspace_id, object_name, ws_type, ...body }) =>
+      query: ({
+        selection_id,
+        workspace_id,
+        object_name,
+        ws_type,
+        description,
+      }) =>
         collectionsService({
           method: 'POST',
           url: encode`/selections/${selection_id}/toset/${workspace_id}/obj/${object_name}/type/${ws_type}`,
-          body: body,
+          body: { description },
           headers: {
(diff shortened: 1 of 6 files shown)


codytodonnell (approved):
This all looks good. The only issue I'm seeing locally is that the scrollbar bugs out in the modal while the match is processing. I'm not sure if this is related to a change you made or not. I could merge your changes into the modal cleanup branch though and handle the issue there.

I also don't have any read-only narratives so I can't confirm that feature. However I am not seeing any narratives in the export menu which could be a good sign.

dauglyon:
During a match or during export? I only changed the export modal so I'm assuming the latter? Can you screencap and share?

dauglyon:
Also just realized I committed with my error image code changes still there ("badtype")

codytodonnell:
> During a match or during export? I only changed the export modal so I'm assuming the latter? Can you screencap and share?

Screenshot 2024-03-11 at 5 13 02 PM

It's in the match modal. The scrollbar expands and detracts while loading

dauglyon:
Hmm yeah, shouldn't be this branch which is causing that. Is main different for you? I can't reproduce

codytodonnell:
I'm seeing it on other branches too so it must be coming from somewhere else.

PR #173, opened 2024-03-18
# URO-188 - add additional search params

eapearson:
# URO-188 - add additional search params

Europa supports a restricted set (“whitelist”) of URL “search parameters”. The “search parameters” are derived from the URL “query component” interpreted in the commonly used format. The DOM API’s [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) serves as the de-facto standard implementation of URL parameters carried in the URL query component, at least for browser-based web apps.

In Europa, a URL with unsupported search parameters results in a removal of the unsupported parameters and some sort of URL rewriting. (How this occurs is out of scope.) For legacy URLs, the result is that the unsupported parameters may be passed to kbase-ui, but they are removed from the visible URL in the location bar. This means that, even if the URL functions correctly the first time, reloading the page, or capturing the url, results in a failure to render the original page view.

To remedy this, given the current design, any search params which utilized by kbase-ui or its plugins should be added to the location which defines the whitelist. This is the file `src/features/paramsSlice.ts`.

This PR adds such support, and also adds a test for file `paramsSlice.ts` which ensures that the newly added parameters are supported.


eapearson:
This is the second in a series of PRs to improve Europa support for kbase-ui.

dauglyon (approved):
Looks great, thanks for catching this. Should be good to merge at your convenience.

PR #174, opened 2024-03-18
add disabled state for the sign-in button [URO-188] 

eapearson:
# URO-189 - disable sign-in button when on sign-in view

This is a relatively minor regression from kbase-ui behavior, extracted from the larger set of improvements to Europa/kbase-ui integration.

During the sign-in process, the sign-in button should not be available, as it will either be non-functional (pressing it on the sign-page is essentially a no-op), or interrupted the sign-in flow (when on the sign-in continuation/confirmation view it would essentially interrupt the sign-in session without cancelling it).

The solution is to teach the sign-in button rendering code to recognize the url paths on which the sign-in button should not be active. There were a couple of tricky bits to this implementation:

- in development, a prefix `/dev/` is for some reason used, which means that url pathname references cannot be absolute, so the pattern’s used do not assume, for instance, that `/legacy/` is the prefix for the pathname, but is merely present somewhere within it.
- the `Link` component used for the sign-in button, provided by “react router”, does not have a disabled state, so the disabled button is replicates the link component, using an actual html `button` element in a disabled state, and with no events attached.

In addition, a test file was added to ensure this behavior was implemented.

### DIFF
--- a/src/app/Routes.tsx
+++ b/src/app/Routes.tsx
@@ -1,4 +1,4 @@
-import { FC, ReactElement } from 'react';
+import { FC, ReactElement, useEffect } from 'react';
 import {
   Navigate,
   Route,
@@ -21,12 +21,14 @@ import {
   detailDataProductPath,
 } from '../features/collections/Collections';
 import {
+  useAppDispatch,
   useAppSelector,
   useFilteredParams,
   usePageTracking,
 } from '../common/hooks';
+import { setLoginControlDisabled } from '../features/layout/layoutSlice';
 
-export const LOGIN_ROUTE = '/legacy/login';
+export const LOGIN_ROUTE = `/${LEGACY_BASE_ROUTE}/login`;
 export const ROOT_REDIRECT_ROUTE = '/narratives';
 
 const Routes: FC = () => {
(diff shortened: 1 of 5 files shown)


dauglyon:
@eapearson I think this may be better (less manual URL rangilng) solved with a wrapper component like `Authed` that sets up a react context to disable login, that way we can define the routes directly in routes.tsx. Tests look great.

Also, /dev/ is set in development to catch cases where we arent handling react-app `PUBLIC_URL` environment variable prefixing manually. Which it seems to have done, but maybe not clearly with regards to reasoning. May be good for us to add this to the README?

Do you mind if I stack a PR on this branch with some suggested changes re: the above?

eapearson:
@dauglyon I'll trust you on the routing business if you think it would be better.

I think the only place I ran into the `/dev/` prefix getting in the way was when proxying to it from kbase-ui. I need to either change it temporarily in Europa to '/' or alter the proxy to use the `/dev/' prefix. It is easiest to just change it in Europa, as it makes it easier to switch back and forth between the dev server and built server. That isn't a big problem, I just need to remember to do it.

And altering this PR is just fine.

eapearson:
@dauglyon This is the other PR, not the final one, but having it done will reduce the size of the final PR.

dauglyon:
@eapearson we will need to make sure the tests pass first, I think I ruined them with my committed changes. I will take a look

eapearson:
@dauglyon All good, I saw that too, that was just a tickle to move this along.

dauglyon:
superseded by #229

PR #178, opened 2024-03-19
update to node 20, which implies npm 10 [URO-191]

eapearson:
This set of changes moves the codebase from requirement to use node 16 to a requirement to use node 20, the current recommended LTS release. Node 16 is out of maintenance and will receive no further updates AFAIK, security or otherwise.

The version chosen is the current release of node 20 at this point in time.

There are, surprisingly, minimal changes to support this. The package lock file, for instance, is unchanged other than the engines property. Tests run fine.

This also allows using node 20 base images, which contain far fewer CVEs than the node 20 ones.


dakotablair (approved):
It's about time! Thanks!

PR #179, opened 2024-03-19
RE2022-307 Filter Contexts

dauglyon:
This PR introduces the use of filter contexts, which are controlled by the "Filter Context Tabs" (All/Matched/Selected). Fixed filter contexts can also be set for dataproducts without matched/selected tabs. FilterContexts are strings of the form `[scope].[mode]` where scope is the "filter source" (eg biolog, genomes, samples), and mode is the view mode for that scope (all/matched/selected). These contexts allow different filters to be applied across different scopes AND across different modes (tabs).

This PR includes the state and queries needed to support filtering in all dataproducts, but only implements filtering in the genome_attribs dp. Implementing samples filtering should be a quick PR after this one.


codytodonnell (commented):
This is looking fantastic. I'm noticing two issues though:

1. The species taxa chart seems to be constantly reloading on both the Taxa Chart page and the Overview page
2. The matches are cleared out when you exit the collection (i.e. go back to the collections browsing page) but if you return to the collection the matching button still shows "Matching by MinHash Homolgy." I might expect the matches to persist until I specifically click "Clear Match" or until I reload the page

codytodonnell:
Two other things I just noticed:
3. Pagination state applies to all three contexts, so if I have lots of matches but only a few selected, I could switch over to the Selected tab and be on a page that doesn't exist. I think it'd be good to automatically reset to page 1 whenever you change tabs.
4. This may just be personal preference but I think it would be smoother if the filters applied to all three contexts. It could get confusing to mentally manage the different sets of filters. The one where it would make sense to have a way to disable the filters is the Selected tab

codytodonnell (approved):

PR #180, opened 2024-03-19
URO-189 (suggestions) add DisabledLogin wrapper element, state.layout.loginControlDisabled

dauglyon:
- Pulls login menu disabled state into redux, and creates a wrapper element for use in Routes.tsx which controls said state.
- Use MUI `Button` with 'link' role in place of existing react-router `Link` (this slightly changes the text style on the button, but I think this is ok, may be worth a check with @codytodonnell)

### DIFF
--- a/src/app/Routes.tsx
+++ b/src/app/Routes.tsx
@@ -1,4 +1,4 @@
-import { FC, ReactElement } from 'react';
+import { FC, ReactElement, useEffect } from 'react';
 import {
   Navigate,
   Route,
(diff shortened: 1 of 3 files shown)


eapearson (commented):
Looks fine to me, but, you know, can't make total sense of it in a diff.
It is nice to have it at the route level as a global concern, and nice to have the more full featured MUI button rather than the react-router link.
I'll need to tweak it a bit later, when merged into the legacy changes. I have a few changes to routing (e.g. handling bare legacy directly and not sending it to kbase-ui, slightly different handling of the fallback)
I might have a comment here or there, but looks fine to me.

dauglyon:
Oh, missed that ^. I will merge the stacked PR I made into this one.

dauglyon:
Cool, I'll go ahead and merge into your branch to keep our PR list uncluttered.

PR #182, opened 2024-03-22
Modify chart layout and polish other UI elements

codytodonnell:
**Chart layout modification**

This moves the scatter chart and histogram to the top of the collection UI (above the table) so that they are easily visible to give users a quick overview of the data. Their height has been shrunken so that the table stays above the fold.

The charts also now have responsive widths and should stack on medium-sized or smaller screens. I had to add a redux state variable `filterPanelOpen` so that the chart components could update themselves and their widths when the filter panel is opened or closed. This is required to keep them responsive.

**Other UI modifications**

This PR also includes several small tweaks to the UI including fixes to number formatting, language clarification in buttons, panel styles, sidebar styles, and match strategy information in the success modal.


src/features/collections/CollectionDetail.tsx
```diff
@@ -216,8 +226,12 @@ export const CollectionDetail = () => {
           <FilterMenu
             collectionId={collection.id}
             anchorEl={filterMenuRef.current}
-            open={filterOpen}
-            onClose={() => setFiltersOpen(false)}
+            open={filterPanelOpen ?? false}
+            onClose={() => {
+              if (collection?.id) {
+                dispatch(setFilterPanelOpen([collection.id, !filterPanelOpen]));
```
dauglyon, line 232: Probably don't need to pass the collectionId, can just put filterPanelOpen in the collectionSlice top level state

dauglyon (commented):
Overall looks good, should be possible to merge my branch on top of this (will be easier than the opposite). I have some concerns about the stability of the chart resize stuff but we can discuss

src/features/collections/data_products/AttribScatter.tsx
```diff
@@ -12,15 +12,15 @@ export const AttribScatter = ({
   xColumn,
   yColumn,
   downsample = 10000,
-  size = [600, 600],
+  size,
```
dauglyon, line 15: write /** doc comment explaining the auto-size situation

dauglyon (approved):

PR #183, opened 2024-03-25
fix categorized filters

dauglyon:


codytodonnell:
This looks good, filters are restored. 

I'm noticing a new issue now where the state of the filter components (e.g. the sliders) doesn't update when switching between contexts when the filter panel remains open. If you close and reopen the panel then the inputs update as expected.

dauglyon:
Hmm, will have to add a context dependency to whatever effect is controlling the state, I'll take a look

src/features/collections/CollectionDetail.tsx
codytodonnell, line 1: Ah interesting way to solve this issue

codytodonnell (approved):
Looks good!

PR #188, opened 2024-03-27
Fix initial axes rendering bug

codytodonnell:


dauglyon (approved):
Oh! Yeah, that'll do it. Thanks for finding the source of that bug.

PR #189, opened 2024-03-27
enable and debounce classification search bar

dauglyon:


codytodonnell (approved):
Quick turnaround! Working well on my end. 

The only issue I'm seeing is that the values between the search bar and the classification input in the filters panel can get out of sync but I think we can live with this issue for now.

PR #191, opened 2024-03-29
RE2022-236: Heatmap label fixes.

dakotablair:
- [x] Heatmap label fixes.
  - [x] Adjust tick font size to improve readability.
  - [x] Fix display of data when the number of rows is less than the page size.
  - [x] Compute row headers, cell values and cell labels at once.
  - [x] Improve code readability.

### DIFF
--- a/src/features/collections/data_products/Biolog.tsx
+++ b/src/features/collections/data_products/Biolog.tsx
@@ -19,7 +19,7 @@ import { useAppDispatch, useBackoffPolling } from '../../../common/hooks';
 import { formatNumber } from '../../../common/utils/stringUtils';
 import { useAppParam } from '../../params/hooks';
 import classes from '../Collections.module.scss';
-import { /* useMatchId, */ useGenerateSelectionId } from '../collectionsSlice';
+import { useGenerateSelectionId } from '../collectionsSlice';
 import { HeatMap, HeatMapCallback, MAX_HEATMAP_PAGE } from './HeatMap';
 
 export const Biolog: FC<{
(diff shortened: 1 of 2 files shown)


src/features/collections/data_products/Biolog.tsx
```diff
@@ -19,7 +19,7 @@ import { useAppDispatch, useBackoffPolling } from '../../../common/hooks';
 import { formatNumber } from '../../../common/utils/stringUtils';
 import { useAppParam } from '../../params/hooks';
 import classes from '../Collections.module.scss';
-import { useGenerateSelectionId } from '../collectionsSlice';
+import { /* useMatchId, */ useGenerateSelectionId } from '../collectionsSlice';
```
dauglyon, line 22: Maybe delete vs comment? Not sure why it's commented

src/features/collections/data_products/HeatMap.tsx
```diff
@@ -394,6 +416,23 @@ export const PlotlyWrapper = ({
     .fill(0)
     .map((z, ix) => ix);
   const otherProps = { ...tooltipProps, visible: tooltipRole };
+  const tickfontSizeCol = (range: number) => {
```
dauglyon, line 419: What is "range" here?

dauglyon: The width? I'm unclear on this function and why the numbers are what they are (if it's just magic looks-good numbers, maybe add a comment?)

dauglyon (approved):
Overall looks good, but I'd clarify the fontsize funcs before merging (see my comment)

src/features/collections/data_products/Biolog.tsx
dakotablair: Ah yes, that was part of the other RE2022-236 work. We'll need `useMatchId` for that, but I'll remove it for the purposes of this PR.

src/features/collections/data_products/HeatMap.tsx
dakotablair: Yes, range is `xmax - xmin` for the column labels and `ymax - ymin` for the row labels. The numbers are indeed magic _looks-good-enough_ numbers, but this also shows how we can update most of the plotly config on render. Doing something like this may make sampling easier than something more fancy. I'll add a comment.

PR #193, opened 2024-04-03
Sample context tabs

dauglyon:


codytodonnell:
I'm running into some weird behavior with the Selected tab on the samples page. Sometimes the count in the tab is showing all the samples but the table is showing them filtered to 2-4. It's hard to reproduce consistently but see if you see the issue by following these steps:

1. Open ENIGMA collection
2. Go to Genome Attributes
3. Select 2 genomes
4. Go to Sample Attributes
5. Click on Selected tab
6. Go back to Genome Attributes
7. Select 2 more genomes
8. Go to Sample Attributes
9. Click on Selected tab

You may have to repeat 6-9 a couple times.

dauglyon:
hmmm, thanks. Ill take a look

codytodonnell (approved):
This is working great for me now. I like you solution linking to the genomes table with the ids filter applied. It might be a little confusing how you lose your existing filters but that's how I would expect it to work in this case. It connects the samples to the genomes right now so I'd say it's a value add.

codytodonnell:
One thing I noticed is that all the genome count links get marked with the :visited state after you click on any of them. Maybe we can disable the :visited state on those links for now.

dauglyon:
@codytodonnell lets run it by the team at standup before I merge, but I imagine they'll defer to you

PR #195, opened 2024-04-04
Fix Filter Slider NaN Bug

dauglyon:
Fixes slider NaN bug by validating inputs before updating slider enables inclusive ranges (can now specify a range with equal min&max)


codytodonnell (approved):
Looks and works great!

PR #196, opened 2024-04-04
fix `Collections search bar doesn’t sync with collections filter`

dauglyon:
fixes bug `Collections search bar doesn’t sync with collections filter`


codytodonnell (approved):
Looks good

PR #204, opened 2024-04-26
RE2022-310 Implement Heatmap filters

dauglyon:
- [x] Add filter menu to Biolog and Microtrait Page
- [x] ensure setting a filter filters the view(s)

This PR also adds support for the "Bool" type filter, but implements it as a range filter (as that's all we have UI for at the moment).

### DIFF
--- a/src/common/api/collectionsApi.ts
+++ b/src/common/api/collectionsApi.ts
@@ -177,7 +177,7 @@ export type ColumnMeta = {
   display_name?: string;
 } & (
   | {
-      type: 'date' | 'int' | 'float';
+      type: 'date' | 'int' | 'float' | 'bool';
       filter_strategy: undefined;
       min_value: number;
       max_value: number;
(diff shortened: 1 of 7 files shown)


src/common/api/collectionsApi.ts
```diff
@@ -177,7 +177,7 @@ export type ColumnMeta = {
   display_name?: string;
 } & (
   | {
-      type: 'date' | 'int' | 'float';
+      type: 'date' | 'int' | 'float' | 'bool';
```
dauglyon, line 180: this was a mistake oops
```suggestion
      type: 'date' | 'int' | 'float';
```

src/features/collections/Filters.tsx
```diff
@@ -280,7 +281,7 @@ export const useContextFilterQueryManagement = (
                 context,
                 column.col_id,
                 {
-                  type: filterMeta.type,
+                  type: 'int',
```
dauglyon, line 284:
```suggestion
                  type: filterMeta.type,
```

PR #205, opened 2024-04-26
RE2022-340 GTDB special casing docs

dauglyon:


src/features/collections/README.md
MrCreosote, line 1: isn't there special casing for the plots as well? I thought there were some other cases but they're not coming to mind

dauglyon: Accidentally missed one related to the plots, adding it. But that seems to be it.

dauglyon: added

MrCreosote: I'm talking abour special casing the axes of the plots, not just special casing them for gtdb. e.g. have the backend specify the plots

dauglyon: Oh, at the moment the plots are "hard coded" with column names (as component parameters) to use for the axes. IIRC we didn't decide to specify those things on the server. Those component parameters are then special cased for GTDB as it has different column names.

dauglyon: Maybe it would be good to also document anywhere column names are used "hardcoded"?

dauglyon: I can think of two off the top of my head: plots, and the classification search box

MrCreosote: Yeah, hardcoded stuff = special casing to my mind

package.json
```diff
@@ -128,7 +128,7 @@
       "lint": {},
       "lint-list-item-indent": "mixed",
       "lint-maximum-line-length": 80,
-      "lint-no-missing-blank-lines": true,
+      "lint-no-missing-blank-lines": false,
```
dauglyon, line 131: this being set to true returns a false positive lint error for nested markdown lists

src/features/collections/README.md
dauglyon: done :)

package.json
dauglyon: @dakotablair you enabled this, so you may want to review its disabling

MrCreosote (commented):
LGTM but @dakotablair should give the final approval

package.json
dauglyon: Hmm, changing the spacing didn't work for me (it then complained about incorrect indent levels)

PR #207, opened 2024-05-02
Some auth cleanup for clarity

dauglyon:


dakotablair (approved):
Looks good to me, I like how you reorganized these tests and improved their clarity.

PR #209, opened 2024-05-10
URO-185: Restored status view inadvertently removed.

dakotablair:
See [2f06f3](https://github.com/kbase/ui/commit/2f06f3).

My bad 😅

### DIFF
--- a/src/app/Routes.tsx
+++ b/src/app/Routes.tsx
@@ -5,7 +5,12 @@ import {
   Routes as RRRoutes,
   useLocation,
 } from 'react-router-dom';
-
+import {
+  CollectionsList,
+  CollectionDetail,
+  detailPath,
+  detailDataProductPath,
+} from '../features/collections/Collections';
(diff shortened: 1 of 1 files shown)


src/app/Routes.tsx
```diff
@@ -36,6 +36,7 @@ const Routes: FC = () => {
   return (
     <RRRoutes>
       <Route path={`${LEGACY_BASE_ROUTE}/*`} element={<Legacy />} />
+      <Route path="/status" element={<Authed element={<Status />} />} />
```
dauglyon, line 39: Does this need to be Authed?

dauglyon (approved):
Left a comment, should remove `Authed` wrapper unless it's strictly needed

src/app/Routes.tsx
dauglyon:
```suggestion
      <Route path="/status" element={<Status />} />
```

PR #213, opened 2024-05-16
Add boolean dropdown filters for heatmap

codytodonnell:
Adds support for boolean filter controls. Uses a dropdown with True, False, or Any as the options where Any is the empty state of the filter. Includes type-support for a new filter type called `'bool'`.

### DIFF
--- a/src/common/api/collectionsApi.ts
+++ b/src/common/api/collectionsApi.ts
@@ -177,7 +177,7 @@ export type ColumnMeta = {
   display_name?: string;
 } & (
   | {
-      type: 'date' | 'int' | 'float' | 'bool';
+      type: 'date' | 'int' | 'float';
       filter_strategy: undefined;
       min_value: number;
       max_value: number;
@@ -196,6 +196,13 @@ export type ColumnMeta = {
       min_value: undefined;
       max_value: undefined;
       enum_values: string[];
+    }
+  | {
+      type: 'bool';
+      filter_strategy: undefined;
+      min_value: undefined;
+      max_value: undefined;
+      enum_values: undefined;
     }
 );
 
--- a/src/common/api/collectionsApi.ts
+++ b/src/common/api/collectionsApi.ts
@@ -177,7 +177,7 @@ export type ColumnMeta = {
   display_name?: string;
 } & (
   | {
-      type: 'date' | 'int' | 'float';
+      type: 'date' | 'int' | 'float' | 'bool';
       filter_strategy: undefined;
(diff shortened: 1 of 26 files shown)


src/common/api/collectionsApi.ts
```diff
@@ -177,7 +177,7 @@ export type ColumnMeta = {
   display_name?: string;
 } & (
   | {
-      type: 'date' | 'int' | 'float';
+      type: 'date' | 'int' | 'float' | 'bool';
```
dauglyon, line 180: unless this causes type issues, I would remove bool from this section. 

```suggestion
      type: 'date' | 'int' | 'float';
```

src/common/components/FilterChip.tsx
```diff
@@ -18,7 +18,7 @@
   ...chipProps
 }: FilterChipProps) => {
   let filterString = '';
-  if (filter.value) {
+  if (filter.value !== undefined && filter.value !== undefined) {
```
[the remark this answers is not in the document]
dauglyon: Presume you meant to check for undefined and null

src/features/collections/CollectionDetail.tsx
```diff
@@ -977,3 +1001,67 @@ const TextFilterControls = ({
...
+  filter: { type: 'bool' };
+}) => {
+  const dispatch = useAppDispatch();
+  const [selectValue, setSelectValue] = useState<string>(() => {
+    if (filter.value === true || filter.value === 1) {
+      return 'true';
+    } else if (filter.value === false || filter.value === 0) {
+      return 'false';
+    } else {
+      return 'any';
```
dauglyon, line 1020: A helpful change!

src/features/collections/CollectionDetail.tsx
```diff
@@ -977,3 +1001,67 @@ const TextFilterControls = ({
...
+        column,
+        { ...filterRef.current, value },
+      ])
+    );
+  }, [selectValue, collectionId, context, column, dispatch]);
+
+  useEffect(() => {
+    //Clear when the filter is cleared
+    // Use 'any' for empty value so the select stays controlled and shows "Any" in the dropdown
+    if (filter.value === undefined || filter.value === null) {
```
dauglyon, line 1053: You could probably check here if the filter value is *different* than filterRef.current.value instead of only handling the any/undefined (cleared) state, but make sure it doesn't cause an infinite loop 😅

dauglyon (approved):

PR #217, opened 2024-06-06
Add log in layout design

codytodonnell:
Added layout/interface pieces for the log in page.

- New `login` directory
- New `LogIn` component and scss module
- New route at `/login`
- Updated UI based on heuristic evaluation: https://docs.google.com/spreadsheets/d/1y9oxlnfG-xS0DBAEAFvZt2q5jLlj1U8z6Nswp_LfYPY/edit?gid=1098038498#gid=1098038498


dauglyon (approved):
Sorry it took me so long, but this should be good to merge! Route shouldn't conflict with anything, so we don't need to feature flag this if you'd like to merge it.

PR #218, opened 2024-06-11
Add new Sign Up layout

codytodonnell:
Add new Sign In pages, components, and UI layout.

- New `signup` route and feature directory
- Three subpages in the sign up flow separated into their own components
  - Sign up provider select
  - Account information
  - KBase use policies


dauglyon:
merged into #229

PR #219, opened 2024-06-18
RE2022-315 Add column select dropdown, no defaults yet

dauglyon:


codytodonnell (approved):
Looks and works good on my end!

PR #220, opened 2024-06-20
URO-211 part A - support for kbase service mocking

eapearson:
This PR extracts changes form the rest of URO-211, which adds support for creating ORCID Links (but deleting, managing, or user profile integration). 

The tests for those changes requires extensive mocking of orcidlink service endpoints. It was then that I discovered that the baseUrl was not specifically configured for the test environment. Rather, without this configuration it defaulted to the production urls - which in turn default to CI. Two sets of tests were actively contacting CI services in their tests and immediately broken when I added support for the test environment baseUrl to orient to http://localhost.

So this set of changes includes:
- setting baseUrl to http://localhost for the node test enviromnet
- improving support for KBase JSON-RPC 1.1 service mocking
- updating two sets of tests to utilize those mocks, rather than contacting CI directly

These changes pave the way for the rest of URO-211, and reduce the # of changes in that future PR.

### DIFF
--- a/src/test/kbaseServiceMock.ts
+++ b/src/test/kbaseServiceMock.ts
@@ -1,0 +1,107 @@
+/**
+ * A testing http server for simulating KBase service endpoints.
+ *
+ * When testing components, apis, or other elements of Europa which depend upon contacting
+ * external resources via http, the recommended approach is to avoid mocking
+ * local code like api clients, but rather to provide an actual api endpoint.
+ *
+ * This implementation uses the "mirage" library, which intercepts fetch
+ * requests.
+ *
+ * One implication of this is that ALL fetch requests will be run against the
+ * mock server. Any tests which inadvertently rely upon contacting CI services
+ * will fail unless support is added herein.
+ *
+ * Note too that the host origin is 'http://localhost'. See
+ * `src/common/api/index.ts` which defines baseUrl for the test environment.
+ */
+import { MockResponseInit } from 'jest-fetch-mock/types';
+import {
+  STATUS_1,
+  VIEW_ASSEMBLY_BRIEF_INFO_1,
+} from './data/narrative_method_store';
+import { KBASEUITEST_PROFILE } from './data/user_profile';
+import { jsonrpc11_resultResponse } from './jsonrpc11';
+
+/**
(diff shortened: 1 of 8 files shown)


eapearson:
@dauglyon would you mind having a look at this? I'm working on cleaning up the follow-up PR, which will be the meat of it and should be ready by the end of the day.

dauglyon:
Hi! Will review today yup. In general looks good. Mocking the requests is a good step, usually I'd recommend a slightly less single-file strategy but given timeline constraints I think someone else can pick up that improvement later on. Otherwise I just still need to do my usual close-read code pedantry.

src/test/kbaseServiceMock.ts
```diff
@@ -0,0 +1,107 @@
...
+import { KBASEUITEST_PROFILE } from './data/user_profile';
+import { jsonrpc11_resultResponse } from './jsonrpc11';
+
+/**
+ * Creates response for fetch mock supplied by jest-fetch-mock.
+ *
+ * Most data used for responses is located in the "data" directory.
+ *
+ */
+export function makeKBaseServices() {
```
dauglyon, line 32: Why "make" and not "mock"?

dauglyon (approved):
Looks good! See prior comment about some potential future changes, otherwise just one question about naming

src/test/kbaseServiceMock.ts
eapearson: just because it is making an instance of kbase service mock endpoints and method implementations. It is making a mock...

eapearson:
> Hi! Will review today yup. In general looks good. Mocking the requests is a good step, usually I'd recommend a slightly less single-file strategy but given timeline constraints I think someone else can pick up that improvement later on. Otherwise I just still need to do my usual close-read code pedantry.

Yes, there are different patterns one could use for setting up mock endpoints. I've found here and in other projects that most test cases are satisfied by one whole, consistent set of mocked endpoints. It creates a mocked-service alternate universe that is consistent between tests. That said, it can be useful to override the service mocks per test, to supply just the necessary mocks per test (gets very verbose for tests, though, so I removed most of those from the second set of changes coming soon), to override a single method mock (e.g. simulate an error in just one method), etc. There are some examples of these techniques in the next set of changes.

PR #221, opened 2024-06-26
implements URO-211 part B - link creation

eapearson:
[ DRAFT ] - still need to create the PR description, observe the GHA outcome, and respond to code quality alerts. Then I'll request a review. Not sure I'll have time to respond to all requests for change before the EOD Friday, but we can try.
The purpose of this PR is to introduce the functionality and user interface to support the creation of an ORCID Link. It builds upon previous work to establish the basic UI components and testing support.

It does not include the following functionality:

- create link from other location (i.e. user profile)
- manage own link
- orcidlink manager manages orcidlink service

The latter two will be implemented in a follow-up PR (or at least branch).

The main area of concern in this PR is probably the approach to interacting with the orcidlink service during linking. Because the linking process is fleeting, and doesn’t really affect application state, a non-RTK client is used in some places. This is also because I didn’t have time to understand RTK well enough to implement this style of query. Anyway, the code for this was extracted from kbase-ui, modified to fit, and (new) tests written.

Another area of note is the usage of “compound components”, in which a stack of related components implements the navigation entry points. It is similar to MVC, in that there is a controller, and there is a view. The controller is responsible for fetching data and providing action functions to the view, and the view is primarily responsible for the ui (although the controller may display loading and error messages). In some cases there is also an “entrypoint” component which is responsible for ensuring that route parameters are populated. The latter is required by various factors, such as the optional nature of route params and search params and the decoupling of app state and component flow (e.g. getting auth state even when behind the Authed component).

I imagine that some of these implementation decisions could be solved by other means.


dauglyon:
superseded by #233

PR #224, opened 2024-07-23
Redesign left and top navbars

codytodonnell:
(WIP) Modify the left and top navbars with UX and aesthetic improvements.

### DIFF
--- a/src/features/layout/TopBar.tsx
+++ b/src/features/layout/TopBar.tsx
@@ -1,31 +1,21 @@
 import {
-  faBars,
-  faEnvelope,
-  faFile,
   faFlask,
   faIdCard,
-  faInfo,
-  faPlus,
-  faQuestion,
   faQuestionCircle,
-  faSearch,
-  faServer,
-  faSignIn,
   faSignOutAlt,
   faSortDown,
   faUser,
   faWrench,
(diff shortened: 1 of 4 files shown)


dauglyon:
What are the implications for documentation here? Off the top of my head, I imagine some docs explain to use the hamburger menu, but perhaps not.

src/features/layout/TopBar.tsx
```diff
@@ -38,14 +28,25 @@ export default function TopBar() {
   return (
     <header className={classes.topbar}>
       <div className={classes.topbar_item}>
-        <HamburgerMenu />
-      </div>
-      <div className={classes.topbar_item}>
+        {/* TODO: replace with recntagular logo without tagline */}
```
dauglyon, line 31: Whats the plan here? Does the asset not exist?

codytodonnell: I was able to find the asset and add it in. I'll remove the TODO.

codytodonnell:
@dauglyon yes I think I'll have to do a sweep of the docs and see where that menu comes up. Otherwise it will just be replacing images.

dauglyon:
Merging per UI meeting conversations

PR #225, opened 2024-08-12
Support Default Columns (last PR for v1.0.10)

dauglyon:
adds default hidden/visible columns, frees Gavin from UI meetings

To compare this and the previous state of things, see the default hidden columns / selected in the dropdown at https://ci-europa.kbase.us/collections/GTDB/genome_attribs vs localhost (fewer visible by default)

### DIFF
--- a/src/common/components/Table.tsx
+++ b/src/common/components/Table.tsx
@@ -1,4 +1,4 @@
-import { useId, useMemo, useState } from 'react';
+import { useEffect, useId, useMemo, useState } from 'react';
 import {
   createColumnHelper,
   ColumnDef,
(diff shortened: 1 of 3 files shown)


codytodonnell:
The column selector and defaults are working well. I think I spotted a small bug where the descending/ascending behavior on the column sorting is flipped. I would expect to see descending values when the arrow points down and ascending values when the arrow points up.

dakotablair (approved):
Looks good!

dauglyon:
> The column selector and defaults are working well. I think I spotted a small bug where the descending/ascending behavior on the column sorting is flipped. I would expect to see descending values when the arrow points down and ascending values when the arrow points up.

Fixed in the above commit.

PR #227, opened 2024-09-24
Fix Backup cookie

dauglyon:
This PR fixes the backup cookie issue. It fixes the cookie configuration, adds tests specific to the backup cookie, and adds a missing condition that obfuscated the configuration error and accidentally worked only on login/logout or when a backup cookie was already present.

### DIFF
--- a/scripts/build_deploy.ts
+++ b/scripts/build_deploy.ts
@@ -16,6 +16,7 @@ interface EnvironmentConfig {
   public_url: string;
   backup_cookie?: {
     name: string;
+    domain: string;
   };
 }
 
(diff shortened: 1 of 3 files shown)


scripts/build_deploy.ts
```diff
@@ -60,6 +61,7 @@ const setEnvironment = (
     REACT_APP_KBASE_DOMAIN: domain,
     REACT_APP_KBASE_LEGACY_DOMAIN: legacy,
     REACT_APP_KBASE_BACKUP_COOKIE_NAME: backupCookie?.name || '',
+    REACT_APP_KBASE_BACKUP_COOKIE_DOMAIN: backupCookie?.domain || '',
```
dauglyon, line 64: the main issue

src/features/auth/authSlice.test.tsx
dauglyon, line 1: Added the ability to test cookies independently, added new backup cookie tests

src/features/auth/hooks.ts
dauglyon, line 1: This should catch a mismatched configuration in the future.

dauglyon (commented):
left some comments explaining which change does what

briehl (approved):
Looks good to me. @dakotablair do you want to chime in?

dakotablair (approved):

PR #228, opened 2024-10-21
Fix initial auth/me call, skip query until token is defined

dauglyon:
Super minor PR to remove a call that always fails (auth/me before we have loaded the login token)

### DIFF
--- a/src/features/auth/hooks.ts
+++ b/src/features/auth/hooks.ts
@@ -14,7 +14,7 @@ import {
 export const useAuthMe = () => {
   const dispatch = useAppDispatch();
   const token = useAppSelector(authToken) || '';
-  const authAPIQuery = getMe.useQuery({ token });
+  const authAPIQuery = getMe.useQuery({ token }, { skip: !token });
   useEffect(() => {
     if (authAPIQuery && authAPIQuery.isSuccess && authAPIQuery.data) {
       const data = authAPIQuery.data;


briehl (commented):
Not sure about the syntax here. I get that this skips the query, what does authAPIQuery get set to? null? undefined?

I think this is fine, mostly just curious.

dauglyon:
@briehl The `skip` optional parameter tells rtk-query not to initiate the request via the useQuery hook until it becomes `false`. Effectively, `fetch` doesn't run, `authAPIQuery.data` remains undefined, and `authAPIQuery.isUninitialized` remains true for any run of the hook which occurs before `skip` is false. After `skip` becomes false, fetch is triggered, `authAPIQuery.isUninitialized` turns false, and after the request response, `authAPIQuery.data` is populated and a rerender is triggered. Docs here: https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching

its basically all a workaround for hooks not being allowed to be called conditionally during a render.

dauglyon:
It may be better to use `skipToken` (another way to do this, slightly better-typed behavior), but so far in europa, we've been using the `skip` param. Switching is easy if we need to in the future.

briehl:
Ok, nice. LGTM if that's been the ongoing pattern through now.

briehl (approved):

PR #235, opened 2025-02-04
Add CDM NavItem and CDM Redirect

dauglyon:
Also adds role-based hide/show to NavItems (only CDM admins will see the button right now), and role-based pseudo-404 routing to the Authed component (to hide this route from normal users in prod)


briehl (approved):
`<insert rubber-stamp noise here>`

PR #239, opened 2025-04-08
Add session cookie to narrative_dev

bio-boris:
See https://kbase.slack.com/archives/C4E7KUGTD/p1744055858948379?thread_ts=1744049701.967629&cid=C4E7KUGTD


briehl (approved):
That just duplicates the other prod environments which also need this. Looks good to me.

PR #250, opened 2025-07-18
Add MFA status indicators to login sessions

dauglyon:
Adds support for the new auth2 mfa feature https://github.com/kbase/auth2/pull/471

- Add mfaAuthenticated field to token type definitions
- Create MfaStatusIndicator component with color-coded icons
- Add MFA Status column to both Current and Other login session tables
- Display green check (MFA used), red check (single factor), or grey check (not supported)
- Include tooltips for each status type
- Only show indicators for Login token types

### DIFF
--- a/src/features/account/LogInSessions.tsx
+++ b/src/features/account/LogInSessions.tsx
@@ -19,6 +19,28 @@ import { Loader } from '../../common/components';
 import { useAppSelector } from '../../common/hooks';
 import { useLogout } from '../login/LogIn';
 
+const MfaStatusIndicator: FC<{ mfaAuthenticated: boolean | null }> = ({ mfaAuthenticated }) => {
+  if (mfaAuthenticated === true) {
+    return (
+      <Tooltip title="MFA used">
+        <FontAwesomeIcon icon={faCheck} style={{ color: 'green' }} />
+      </Tooltip>
+    );
+  } else if (mfaAuthenticated === false) {
+    return (
+      <Tooltip title="Single factor">
+        <FontAwesomeIcon icon={faCheck} style={{ color: 'red' }} />
+      </Tooltip>
+    );
+  } else {
+    return (
+      <Tooltip title="MFA not supported">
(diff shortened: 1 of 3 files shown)


src/features/account/LogInSessions.tsx
```diff
@@ -19,6 +19,30 @@ import { Loader } from '../../common/components';
...
+const MfaStatusIndicator: FC<{ mfaAuthenticated: boolean | null }> = ({
+  mfaAuthenticated,
+}) => {
+  if (mfaAuthenticated === true) {
+    return (
+      <Tooltip title="MFA used">
+        <FontAwesomeIcon icon={faCheck} style={{ color: 'green' }} />
+      </Tooltip>
+    );
+  } else if (mfaAuthenticated === false) {
```
ialarmedalien, line 31: no need to use `else` if the previous statement was a `return`

src/features/account/LogInSessions.tsx
```diff
@@ -19,6 +19,30 @@ import { Loader } from '../../common/components';
...
+const MfaStatusIndicator: FC<{ mfa: 'USED' | 'NOT_USED' | 'UNKNOWN' }> = ({
+  mfa,
+}) => {
+  if (mfa === 'USED') {
+    return (
+      <Tooltip title="MFA used">
+        <FontAwesomeIcon icon={faCheck} style={{ color: 'green' }} />
+      </Tooltip>
+    );
+  } else if (mfa === 'NOT_USED') {
```
ialarmedalien, line 31: no need for `else` after a `return`

https://eslint.org/docs/latest/rules/no-else-return

why I started following this rule:

https://docs.astral.sh/ruff/rules/superfluous-else-return/

dauglyon: I feel it makes it less readable... else makes it clear it's exclusive to the blocks above, especially for multiple if-else-ifs

dauglyon: ie I agree it is "textually" (in execution) superfluous but I think it is sub-textually substantial

PR #251, opened 2025-12-04
Add login/signup legacy redirects, add logout path

dauglyon:


briehl (approved):
Looks good. Does the logout() function do anything that can fail? should it get wrapped in a try/catch?

PR #258, opened 2026-05-29
Signup fixes: username validation, ORCID pre-fill, and CDM sidebar role gate

dauglyon:
## Summary

Three small fixes around signup and the authenticated sidebar.

- **Signup username validation**: Mirror the [`kbase/auth2` `NewUserName`](https://github.com/kbase/auth2/blob/main/src/main/java/us/kbase/auth2/lib/NewUserName.java) rules on the signup form: must start with a lowercase letter, only `[a-z0-9_]`, no repeating or trailing underscores, ≤100 chars. The availability check previously compared `availablename` to `username.toLowerCase()`, so inputs like `John` passed the frontend check and were then rejected by the backend. Special-character inputs (`John.Doe`, `bad-user`, …) reported "Username is not available — Suggested: X" which was actually a format problem, not a collision. Now the form shows a specific format error and submit stays blocked until the input matches the backend rules.
- **Skip `user<N>` pre-fill on ORCID signup**: ORCID's provider-supplied username is the numeric ORCID iD (e.g. `0000-0002-1825-0097`). auth2's `NewUserName.sanitizeName` strips all of that to empty and `getAvailableUserName` falls back to the literal `user<N>` (see `Authentication.java:1185-1196` and `DEFAULT_SUGGESTED_USER_NAME`). The signup form was pre-populating the username field with `user1`/`user2`/etc. for every ORCID signup. Leave the field blank when the provider is `OrcID` so the user picks their own. Display name and email pre-fill from ORCID are unchanged.
- **CDM sidebar role gate**: The CDM nav item was gated on `CDM_JUPYTERHUB_ADMIN` (admin only). Per the BERDL platform docs, `BERDL_USER` is the access role; `CDM_JUPYTERHUB_ADMIN` is a separate admin role for approving access requests. Switch the gate so users with `BERDL_USER` see the link.

## Test plan

Signup username validation
- [ ] Signup with a username containing an uppercase letter (e.g. `John`) — submit stays blocked, format error shown.
- [ ] Signup with `bad-user`, `bad__user`, `baduser_`, `1baduser` — submit stays blocked with format error.
- [ ] Signup with a valid lowercased username (e.g. `testuser`) — submit proceeds to step 3.
- [ ] Username `>` 100 chars shows the "must be at most 100 characters" error.

ORCID signup pre-fill
- [ ] Start signup via ORCID — username field is empty (not `user1`).
- [ ] Start signup via Google/Globus/another provider — username field is still pre-filled from `availablename`.
- [ ] Display name and email are still pre-filled from the ORCID identity.

CDM sidebar gate
- [ ] Log in as a user with `BERDL_USER` but no admin role — CDM link visible in sidebar.
- [ ] Log in as a user without `BERDL_USER` — CDM link not visible.

Unit tests
- [ ] `npm test -- --testPathPattern='src/features/signup/'` — all 22 tests pass (includes 6 new parameterized format-error cases and 3 new `setLoginData` cases).

### DIFF
--- a/src/features/layout/LeftNavBar.tsx
+++ b/src/features/layout/LeftNavBar.tsx
@@ -72,7 +72,7 @@ const LeftNavBar: FC = () => {
           icon={faDatabase}
           badge={'alpha'}
           badgeColor={'warning'}
-          requiredRole="CDM_JUPYTERHUB_ADMIN"
+          requiredRole="BERDL_USER"
         />
       </ul>
       <ul className={classes.nav_list}>
--- a/src/features/signup/AccountInformation.test.tsx
+++ b/src/features/signup/AccountInformation.test.tsx
@@ -135,4 +135,72 @@ describe('AccountInformation', () => {
 
     expect(mockNavigate).toHaveBeenCalledWith('/signup/3');
   });
+
+  test.each([
+    ['uppercase letters', 'BadUser'],
+    ['a leading digit', '1baduser'],
+    ['a hyphen', 'bad-user'],
+    ['a period', 'bad.user'],
+    ['repeating underscores', 'bad__user'],
+    ['a trailing underscore', 'baduser_'],
+  ])(
+    'blocks submission and shows format error for username with %s',
+    async (_label, badName) => {
+      const store = createTestStore();
+      store.dispatch(
+        setLoginData({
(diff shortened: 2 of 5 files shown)


briehl (approved):
Looks good!

